PR #6, opened 2023-05-16
Add postgres. Closes #5

### LINKED ISSUE #5: Connect to postgres
> The following secret was added to spark:
  DATABASE_URL=postgres://...

juliangruber:


bin/spark.js
```diff
@@ -1,8 +1,14 @@
 import http from 'node:http'
 import { once } from 'node:events'
-import { handle } from '../index.js'
+import { createHandler } from '../index.js'
+import pg from 'pg'
 
-const server = http.createServer(handle)
-server.listen(8080)
+const { PORT = 8080, DATABASE_URL } = process.env
```
bajtos, line 6: +1 for using `process.env.PORT` and `process.env.DATABASE_URL` 👍🏻

bin/spark.js
```diff
@@ -1,8 +1,14 @@
...
+
+const client = new pg.Client({ connectionString: DATABASE_URL })
+await client.connect()
```
bajtos, line 9: I think we should use a connection pool, otherwise we cannot handle multiple incoming requests concurrently.

See https://node-postgres.com/apis/pool

juliangruber: good call! https://github.com/filecoin-station/spark-api/pull/9

Issue #1, opened 2023-05-16
Job Scheduling


juliangruber:
- Web2-style, let’s get to rewards as quickly as possible.
- A cloud-based Orchestrator service that assigns checker jobs to Station instances. In the future, we should replace this with a smart contract.
- **The simplest job selection possible:** embed a hard-coded list of `(CID, providerEndpoint, protocol)` into the deployed Orchestrator service and pick a random item whenever we need to schedule a job. (We will improve this in the following iterations.)
    - We can get some CIDs from Saturn folks or Rhea/Lassie logs.
    - We find list of peers and protocols for a given CID using the public API provided by `cid.contact`, see [https://docs.cid.contact/query-and-retrieve/querying-indexer-provider](https://docs.cid.contact/query-and-retrieve/querying-indexer-provider)
- The API implementation writes the randomly selected job into our internal database, together with a timestamp. It generates a unique `JobId` field, stores it in the DB and includes it in the response.
- Let’s use [Fly.io](http://fly.io/) because it offers both Server hosting and DB hosting.
- We should implement schema migrations using an automated tool (e.g. postgrator or something else).
- The Station/Zinnia team will operate this orchestrator.
- Setup CI, tests, linting. Setup automatic deployments (CD) on push or tag.

Estimated at 5 days of work

PR #7, opened 2023-05-17
add `PUT /retrievals` implementation. #1

juliangruber:


README.md
```diff
@@ -5,14 +5,25 @@
 
 ## Routes
 
-### `GET /retrieval`
+### `PUT /retrieval`
+
+Start a new retrieval.
```
bajtos, line 10: When I see `PUT /retrieval`, I expect create-or-update semantics - the server should create a new resource or update an existing one, typically using some sort of an unique id.

Quoting from https://restfulapi.net/http-methods/#put

> Use PUT APIs primarily to update an existing resource (if the resource does not exist, then API may decide to create a new resource or not).
>
> If the request passes through a cache and the Request-URI identifies one or more currently cached entities, those entries SHOULD be treated as stale. Responses to PUT method are not cacheable.

IMO, this endpoint is creating (generating) a new retrieval job definition, therefore it's better to use the verb/method `POST`.

Quoting from https://restfulapi.net/http-methods/#post
> Use POST APIs to create new subordinate resources, e.g., a file is subordinate to a directory containing it or a row is subordinate to a database table.
> 
> When talking strictly about REST, POST methods are used to create a new resource into the collection of resources.
> 
> Responses to this method are not cacheable unless the response includes appropriate [Cache-Control](https://en.wikipedia.org/wiki/Web_cache#Cache_control) or [Expires](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html) header fields.
> 
> Please note that POST is neither safe nor idempotent, and invoking two identical POST requests will result in two different resources containing the same information (except resource ids).

I think we should also distinguish between retrieval job definition (what to retrieve) and retrieval report (whether we could download the content and telemetry data about the operation), and use different URLs (endpoint paths) for that. 

For example:

- `POST /retrieval-jobs` - create a new job
- `POST /retrieval-reports` - report a job outcome

This makes the API open to extensions, for example:

- `GET /retrieval-jobs?limit=10` - list 10 most recent retrieval jobs
- `GET /retrieval-jobs/123` - get details about the retrieval job with id 123
- `GET /retrieval-reports?limit=10` - list 10 most recent retrieval reports
- `GET /retrieval-reports/56` - get details about the retrieval report 56

I think we are not likely to implement these endpoints, I am listing them here to give more context to my suggestion.

README.md
```diff
@@ -5,14 +5,25 @@
...
+```typescript
+{
+  id: String,
+  cid: String,
+  providerAddress: String,
+  protocol: 'graphsync'|'bitswap'
+}
 ```
 
+### `POST /retrieval/:id`
```
bajtos, line 23: See my comment above.

I am not sure if the job id should be in the endpoint path. My personal preference is to put it into the request payload.

In REST APIs, ids in paths are usually used to identify the resource we want to interact with.

When creating a new retrieval report, the path should contain the report id, not the job id.

Maybe I have a different mental model of the data schema you want to implement?

README.md
```diff
@@ -5,14 +5,25 @@
...
+
+_TODO_
+
 ## Development
 
 1. Set up [PostgreSQL](https://www.postgresql.org/) with default settings:
```
bajtos, line 29: >  - Port: 5432
>  - User: _your user name_
>  - Password: _blank_
>  - Database: _same as user name_

This comment is probably out of scope of this PR, but I am not very comfortable with this setup.

What is "my username" - is it the same as printed by `echo $USER`? I want to use a different username & database for each project. I'd also like to us a non-empty password.

What privileges must the DB user have? Ideally, I want to use a project-specific user that has access to the project-specific database only. That means that as part of the dev setup, I need to create the database & the user on my PG server. 

It would be great to have a script I can run to set everything up for me.

Alternatively, we can run PG via Docker. That saves us from a ton of complexity. For example, to nuke the DB and start from scratch, we can simply delete the old Docker container and create a new one.

In the past, I was using npm package scripts like these:

```json
{
  "scripts": {
    "db:start": "docker run --name spark-db -p 5432:5432 -e POSTGRES_PASSWORD=spark-space-traveller -d ghcr.io/baosystems/postgis:13-3.1 -c shared_preload_libraries='pg_stat_statements'",
    "db:migrate": "node -r dotenv/config db/migrate.js && db/setup-test-db.js",
    "db:setup-test": "db/setup-test-db.js",
    "db:destroy": "docker rm -fv spark-db",
    "db:reset": "pnpm run db:destroy && pnpm run db:start",
  }
}
```

In this setup, I used two databases - one for development and debugging, the other for testing. That way, `npm test` did not interfere with the data seen and written by `npm start`.

migrations/002.undo.retrievals.sql
bajtos, line 1: Do you expect we will be rolling back migrations? I think we should create a new `do` migration if we ever need to roll back an older migration, and put the roll back code there. In other words, I think we don't have to create `undo` migrations.

test/test.js
```diff
@@ -33,11 +33,29 @@ describe('Routes', () => {
...
+        const res = await fetch(`${spark}/retrieval`, { method: 'PUT' })
+        assert.strictEqual(res.status, 200)
+        const body = await res.json()
+        if (!lastCid) {
+          lastCid = body.cid
+        } else if (lastCid !== body.cid) {
+          return
+        }
+      }
+      throw new Error('All CIDs were the same')
```
bajtos, line 58: 1. I find it quite challenging to understand how exactly the assertion works in this test.

2. The error message seems misleading - we don't know if _ALL_ CIDs are the same, only that _SOME TWO CONSECUTIVE_ CIDs are the same.

3. You are running this loop 100 times, but the database has only 2 rows to choose from. I don't see how we can generate 100 unique CIDs from those 2 rows? Maybe the test passes because the server returns CID1 - CID2 - CID1 - CID2 and we never get two consecutive responses with the same CID. But that's not really correct, right?

I am proposing to make the following changes:
1. Run the loop fewer times. If we have only 2 records then we must run it twice. With more records in the DB, I would choose 10 iterations.
2. Collect all CIDs into an array.
3. At the end, assert that this array contains unique items only. If the assertion fails, then print the entire array to make it easier to understand what went wrong.
  
   To detect repeated items, we can e.g. sort the array and then check for duplicates using your algorithm.

bajtos (changes requested):

test/test.js
juliangruber: It tries 100 attempts to find two different CIDs. I thought the iteration count of 100 is safe enough to not have to think about it. We can also lower it to 50 or less if you prefer?

>  I don't see how we can generate 100 unique CIDs from those 2 rows?

We are never generating CIDs, I don't think we could? There is a fixed list of CIDs, stored in `retrieval_templates`. On every request, we are generating a `retrieval`, based on one of the `retrieval_templates`. This test ensures that `retrieval_templates` are chosen at random. Should I add a comment to make this more clear?

migrations/002.undo.retrievals.sql
juliangruber: Absolutely: If we need to roll back a deployment, the flow is 1) `git revert` 2) `git push`. This works only if the `git revert` will put the repository into a state where the last migration will be undone. I don't think writing a new `do` migration in order to roll back, and then reverting the code on top, will scale.

README.md
juliangruber: I think the difference is that you're thinking about `retrieval job` and `retrieval report`, while I'm only thinking about one object `retrieval`, which gets updated as we go.

I like having just one object for a retrieval. You don't need to join when querying, and also job+report have a 1:1 relationship, so we don't need flexibility gained by decoupling the data.

README.md
juliangruber: See comment below for discussion of the schema

README.md
bajtos: Fair enough. 

It's possible for multiple clients to submit retrieval data for the same job (retrieval). I think we should record such attempts and use the data to detect fraudulent Station operators. I guess your schema does not prevent us from doing so - the first "submit" request will update the database, and subsequent requests will detect that the report was already submitted and record the fraudulent attempt elsewhere.

In that case I am proposing:

(1)
`POST /retrievals` to create a new retrieval


(2)
`PATCH /retrievals/:id` to report the retrieval outcome. Why `PATCH`? Because `PUT` is usually used to update all data fields of a resource. 

https://restfulapi.net/http-methods/#patch
> HTTP PATCH requests are to make a partial update on a resource.
>
> If you see PUT requests modify a resource entity too. So to make it more precise – the PATCH method is the correct choice for partially updating an existing resource, and you should only use PUT if you’re replacing a resource in its entirety.

TBH, I think it's better to abandon the REST design philosophy for this second endpoint and use the RPC style instead:

`POST /submit-retrieval-report` with all data provided in the request body.

README.md
bajtos: > I can see that on another page (https://postgresapp.com/) they put "your system user name" instead of "your user name". Do you think that's better?

Yes, I think "your system user name" is more clear.

> For that you can set the environment variable DATABASE_URL, as is done in deployments. Do you think that's good enough for your dev flow? If so, shall we document it?

That should work, although I will probably need to figure out how to set that env var for spark-api project only.

Yes please, let's mention this env var in the documentation for developers.

How do you feel about using `node -r dotenv/config` so that I can put `DATABASE_URL` into `.env` file?

> I believe db user management is out of scope for this project. The user needs to be able to do anything it wants with the given database. It doesn't need to create users or perform other tasks. Shall we document that?

Sure, I agree DB management is out of scope of this project. It was not clear to me from the docs what I need to do to setup the DB for spark-api.

I think the docs should say that the user and the database must already exist in the DB server and the user must have full access to this particular database, especially permissions to create and alter tables. WDYT?

> > It would be great to have a script I can run to set everything up for me.
>
> We can add that if you want 👍 As long as we don't make it a requirement, I don't want to have to run docker.

We can support both ways:
- A script that runs as `root` or `admin` and sets up everything needed on the local PG server.
- Another script to setup the DB via Docker - this should be pretty much a one-liner.

Let's not worry about the Docker workflow for now. I'll figure it out when I need to run this project locally.

README.md
juliangruber: Aaaah right, POST and PATCH is the right pair 👍 

> TBH, I think it's better to abandon the REST design philosophy for this second endpoint and use the RPC style instead:

Why?

migrations/002.undo.retrievals.sql
bajtos: Fair enough 👍🏻 I didn't consider that we may need to both rollback the database and revert code via git.

test/test.js
bajtos: > It tries 100 attempts to find two different CIDs. I thought the iteration count of 100 is safe enough to not have to think about it. We can also lower it to 50 or less if you prefer?
> 
> > I don't see how we can generate 100 unique CIDs from those 2 rows?
> 
> We are never generating CIDs, I don't think we could? There is a fixed list of CIDs, stored in `retrieval_templates`. On every request, we are generating a `retrieval`, based on one of the `retrieval_templates`. This test ensures that `retrieval_templates` are chosen at random. Should I add a comment to make this more clear?

Sorry for not being clear. By generating CIDs I meant generating retrieval jobs by picking random CIDs.

I see that you have another PR to add more than 2 CIDs to pick from, that resolves one of my concerns.

I am still confused about how is your test supposed to work.
- You store the first CID fetched inside `lastCid` (how confusing!) and then run another requests, up to 99.
- After the second request, if `lastCid !== body.cid` then you `return`, therefore terminating the entire test with success.
- If the second request returns the same CID as the one before, then you keep trying.

I think I am starting to understand it now, but it took me a while. How about rewriting the test along the following lines?

```js
const pickRandomCid = async () {
  const res = await fetch(`${spark}/retrieval`, { method: 'POST' })
  assert.strictEqual(res.status, 200)
  const body = await res.json()
  return body.cid
}

const firstCid = await pickRandomCid()

for (let i = 0; i < 100; i++) {
  let anotherCid = await pickRandomCid()
  if (anotherCid !== firstCid) {
    // The service seems to return different CIDs 
    // for different requests - the test passed.
    return
  }
}

throw new Error(`All requests returned the same CID: ${firstCid}.`)
```

README.md
juliangruber: > How do you feel about using node -r dotenv/config so that I can put DATABASE_URL into .env file?

Sgtm! However I would prefer to use dotenv as a lib

Cool, addressed all remaining points in https://github.com/filecoin-station/spark-api/pull/10

test/test.js
juliangruber: This was really helpful to read! Thank you for sharing. I refactored the test very similar to what you proposed 🙏

README.md
bajtos: > Aaaah right, POST and PATCH is the right pair 👍

👌🏻 

> > TBH, I think it's better to abandon the REST design philosophy for this second endpoint and use the RPC style instead:
> 
> Why?

IMO, we don't want clients (the Spark module) to change a subset of `retrieval` properties arbitrarily, which is my understanding of PATCH semantics. Instead, we want them to submit a retrieval result with metrics, which is a strictly defined RPC-like operation with a list of required request fields.

Anyhow, I don't hold this opinion too strongly. I can live with `PATCH` too.

juliangruber: Implementing PATCH doesn't mean you're allowed to change all fields, it can still only be just the ones we need.

bajtos (dismissed):

PR #12, opened 2023-05-17
add ingestion. closes #2

### LINKED ISSUE #2: Ingester
- Add a new HTTP route to our backend monolith app to receive reports about completed jobs. This will share the same DB with Orchestrator/Job Scheduler.
- When Ingester receives a job report from a Station, it attaches the current timestamp to the record stored. This will help us troubleshoot and may be useful later for fraud detection.
  - The JobId in the report allows us to verify that Stations performed jobs that our Orchestrator scheduled, as the job report must provide a valid JobId.

Estimate: 2 days

juliangruber:
closes #2


README.md
```diff
@@ -22,7 +22,22 @@ Response:
 
 ### `PATCH /retrievals/:id`
 
-_TODO_
+Parameters:
+- `id`: Request ID
```
bajtos, line 26:
```suggestion
- `id`: Retrieval ID as returned by `POST /retrievals`
```

index.js
```diff
@@ -1,11 +1,14 @@
...
+import getRawBody from 'raw-body'
+import assert from 'http-assert'
 
 export const createHandler = async (client) => {
   await migrate(client)
 
   const handler = async (req, res) => {
-    if (req.url === '/retrievals' && req.method === 'POST') {
+    const segs = req.url.split('/').filter(Boolean)
+    if (segs[0] === 'retrievals' && req.method === 'POST') {
```
bajtos, line 11: This is a subtle behaviour change. With your change in place, we start handling all paths starting with `/retrievals`, e.g. `/retrievals/foobar`.

A possible fix:

```suggestion
    if (segs[0] === 'retrievals' && segs.length == 1 && req.method === 'POST') {
```

I wonder how far we can get with this bare-bone DYI approach before having to switch to a proper web framework like Fastify anyways. 🤔

index.js
```diff
@@ -26,16 +29,41 @@ export const createHandler = async (client) => {
         providerAddress: retrievalTemplate.provider_address,
         protocol: retrievalTemplate.protocol
       })
-      return
+    } else if (segs[0] === 'retrievals' && req.method === 'PATCH') {
+      const retrievalId = Number(segs[1])
+      assert(!Number.isNaN(retrievalId), 400, 'Invalid Retrieval ID')
+      const body = await getRawBody(req, { limit: '100kb' })
```
bajtos, line 35: +1 for remember to set a limit on the incoming request body

test/test.js
```diff
@@ -62,4 +62,88 @@ describe('Routes', () => {
...
+          },
+          body: JSON.stringify({
+            success: true
+          })
+        }
+      )
+      assert.strictEqual(res.status, 404)
+      assert.strictEqual(await res.text(), 'Retrieval Not Found')
+    })
+  })
```
bajtos, line 148: Is it worth adding a test for request bodies larger than our limit?

bajtos (approved):

bajtos (approved):

index.js
juliangruber: That changed yeah, but does it matter?

> I wonder how far we can get with this bare-bone DYI approach before having to switch to a proper web framework like Fastify anyways. 🤔

Let's go there when we need to. We can also try some more lightweight approaches first, like https://github.com/pillarjs/path-to-regexp or writing regexps directly.

juliangruber: At this stage I would like to keep complexity low and movability high, we can prevent things like `POST /retrievals/mistake` further down the road.

juliangruber: For reference, I built the electron update server this way too https://github.com/electron/update.electronjs.org/blob/main/src/updates.js#L62-L84, and that hasn't needed changes related to the routing layer for 5 years

test/test.js
juliangruber: good idea!

juliangruber:
@bajtos hindsight please

PR #14, opened 2023-05-24
Add submit wallet address

juliangruber:
For https://github.com/filecoin-station/spark/issues/1

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,8 @@ Body:
 
 ```typescript
 {
-  success: Boolean
+  success: Boolean,
+  walletAddress: String
 }
 ```
 
--- a/index.js
+++ b/index.js
@@ -51,7 +51,7 @@ export const createHandler = async (client) => {
         SET finished_at = NOW(),
             success = $2,
             wallet_address = $3
-        WHERE id = $1
+        WHERE id = $1 AND success IS NULL
         RETURNING id
       `, [
         retrievalId,
--- a/index.js
+++ b/index.js
@@ -33,16 +33,30 @@ export const createHandler = async (client) => {
       const retrievalId = Number(segs[1])
       assert(!Number.isNaN(retrievalId), 400, 'Invalid Retrieval ID')
       const body = await getRawBody(req, { limit: '100kb' })
-      const { success } = JSON.parse(body)
+      const { success, walletAddress } = JSON.parse(body)
+      assert.strictEqual(
+        typeof success,
+        'boolean',
+        400,
+        'boolean .success required'
+      )
+      assert.strictEqual(
+        typeof walletAddress,
+        'string',
+        400,
+        'string .walletAddress required'
+      )
       const { rows } = await client.query(`
         UPDATE retrievals
         SET finished_at = NOW(),
-            success = $1
-        WHERE id = $2
+            success = $2,
+            wallet_address = $3
+        WHERE id = $1
         RETURNING id
       `, [
+        retrievalId,
         success,
-        retrievalId
+        walletAddress
       ])
       assert(rows.length > 0, 404, 'Retrieval Not Found')
       res.end('OK')
--- a/test/test.js
+++ b/test/test.js
@@ -85,19 +85,21 @@ describe('Routes', () => {
             'Content-Type': 'application/json'
           },
           body: JSON.stringify({
-            success: true
+            success: true,
+            walletAddress: 'abc'
           })
         }
       )
       assert.strictEqual(updateRequest.status, 200)
       const { rows: [updatedRetrievalRow] } = await client.query(`
-        SELECT success
+        SELECT *
         FROM retrievals
         WHERE id = $1
       `, [
         retrievalId
       ])
       assert.strictEqual(updatedRetrievalRow.success, true)
+      assert.strictEqual(updatedRetrievalRow.wallet_address, 'abc')
     })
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -29,7 +29,8 @@ Body:
 
 ```typescript
 {
-  success: Boolean
+  success: Boolean,
+  walletAddress: String
```
bajtos, line 33: Uh oh, when reviewing only this change, I find it confusing to see `success` in the request body. This field is usually returned by the server, so I had to double check that this part of our docs is describing the request body.

How do you feel about renaming `success` to something more descriptive, e.g. `retrievalSucceeded`?

I am fine to leave that out of this PR, just pointing out my surprise.

index.js
```diff
@@ -33,16 +33,30 @@ export const createHandler = async (client) => {
...
-        WHERE id = $2
+            success = $2,
+            wallet_address = $3
+        WHERE id = $1
         RETURNING id
       `, [
+        retrievalId,
         success,
-        retrievalId
+        walletAddress
```
bajtos, line 59: What is our plan for rejecting duplicate job-result submissions? Again, out of scope of this PR, just want to make sure we don't forget to handle this particular edge case (and an attack vector).

test/test.js
```diff
@@ -85,19 +85,21 @@ describe('Routes', () => {
             'Content-Type': 'application/json'
           },
           body: JSON.stringify({
-            success: true
+            success: true,
+            walletAddress: 'abc'
```
bajtos, line 89: Can we use test data that at least resembles FIL addresses? When I encounter the value `abc`, I'll have no idea what is it connected to.

For example, we can use `f1abc` or `f1test` instead. 

I think we will need to switch to real FIL addresses in the future and reject HTTP requests with invalid wallet address. To prepare for that, it would be great to extract this dummy test wallet address into a shared constant used by all tests.

bajtos (dismissed):
LGTM 👍🏻 

I have few comments to consider, feel free to address them in follow-up PRs.

test/test.js
juliangruber: good idea!

index.js
juliangruber: Good catch, I forgot that during this PR. Fixed

README.md
juliangruber: > Uh oh, when reviewing only this change, I find it confusing to see success in the request body. This field is usually returned by the server, so I had to double check that this part of our docs is describing the request body.

Interesting, I hadn't thought about it this way at all. I guess to me `success` is _not_ a typical response field, because you should use HTTP code 200 to mark a success.

I was halfway through changing it to `retrievalSucceeded`, but then realized an issue: If you call a field "retrievalSucceeded' when updating a retrieval, you must prefix all fields with "retrieval". Otherwise, you get inconsistent naming. I'm therefore going to keep it as is for now, let's reevaluate in the future.

test/test.js
```diff
@@ -169,5 +169,42 @@ describe('Routes', () => {
...
+      {
+        const updateRequest = await fetch(
+          `${spark}/retrievals/${retrievalId}`,
+          {
+            method: 'PATCH',
+            headers: { 'Content-Type': 'application/json' },
+            body: JSON.stringify({ success: false, walletAddress })
+          }
+        )
+        assert.strictEqual(updateRequest.status, 404)
```
bajtos, line 198: I think the status 404 may complicate troubleshooting errors, because we cannot tell whether there is no such retrieval job with `retrievalId` or whether the job exists but the result was already submitted.

I am proposing to return [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) or [409 Conflict](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) and include a descriptive error message in the response body.

It's fine to make this change in a follow-up PR.

bajtos (approved):

test/test.js
juliangruber: The update fails if we find no row with given id or without a success state. Currently we have no way to tell (without changing the query) whether it failed for the one reason or the other. Can you think of a simple way to change this?

I'm going to merge for now, let's continue discussing and I'll follow up with a PR for this

bajtos: > Currently we have no way to tell (without changing the query) whether it failed for the one reason or the other. Can you think of a simple way to change this?

I think we will need two queries, unless we can find some advanced SQL feature that will give us the behaviour we need.

bajtos: We can still make it atomic.
1. The first query finds the row by id. If the row does not exist, we return 404.
2. The second query updates the row if no result was recorded yet and returns a flag to indicate whether the query updated something or nothing.

There would be a race condition of somebody deleted the row between the first and the second query, in which case we would incorrectly say that the result was already reported instead of 404 not found. However, we are not deleting rows, so this situation is not going to happen.

bajtos: On the second thought, maybe we can do this the other way around and optimise for the happy path:

1. Run your current query. If it succeeds then all is good.
2. If we don't update anything, then run the second query to see if the retrieval id is invalid or we already have a result recorded.

WDYT?

juliangruber: Ah nice I like that optimization for the happy path! Will follow up :)

juliangruber: https://github.com/filecoin-station/spark-api/pull/17

PR #16, opened 2023-05-28
refactor `handler` / `errorHandler`

juliangruber:
🧹

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,46 +6,53 @@ import assert from 'http-assert'
 const handler = async (req, res, client) => {
   const segs = req.url.split('/').filter(Boolean)
   if (segs[0] === 'retrievals' && req.method === 'POST') {
-    // TODO: Consolidate to one query
-    const { rows: [retrievalTemplate] } = await client.query(`
-      SELECT id, cid, provider_address, protocol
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -3,67 +3,68 @@ import { migrate } from './lib/migrate.js'
...
+    `, [
+      success,
+      retrievalId
+    ])
+    assert(rows.length > 0, 404, 'Retrieval Not Found')
+    res.end('OK')
+  } else {
+    res.end('Hello World!')
+  }
+}
```
bajtos, line 49: While you are refactoring this file, would you mind extracting route handlers into individual functions?

```js
const handler = async (req, res, client) => {
const segs = req.url.split('/').filter(Boolean)
if (segs[0] === 'retrievals' && req.method === 'POST') {
  createNewRetrieval(req, res, client);
} else if (segs[0] === 'retrievals' && req.method === 'PATCH') {
  patchRetrievalById(Number(segs[1]), req, res, client);
} else { 
  res.end('Hello World!');
}
```

bajtos (approved):
Nice improvement 👍🏻

bajtos (commented):
👏🏻

PR #18, opened 2023-05-30
split up retrieval schema

juliangruber:
@bajtos your initial intuition was right (or you already knew): It's better to decouple the retrieval request from the retrieval result.

In this particular case, it's particularly better because it helps with DB validations. Before, a `retrieval` row could have the optional fields `success` and `finished_at`, but the DB couldn't validate that either none or both have to be set. This is why I've updated the schema to a 1:1 relationship between `retrieval` and `retrieval_success`. This also made checking whether a retrieval result already exists or the retrieval doesn't exist trivial.

Please tell me what you think about this way of handling validations. I really like it because it makes the DB the one source of truth, and it's way better at that than nodejs logic can ever be.

I've also changed the migrations naming scheme. We were previously using sequential numbers, which caused issues when working on different PRs: Both PR can have a migration with the same number, which then leads to a conflict. Timestamps should be a safer way to deal with this.

After this PR, adding more columns for the retrieval results will be easy: Just extend the schema and the insert query. No new validations need to be written.


index.js
```diff
@@ -37,35 +38,31 @@ const createRetrieval = async (res, client) => {
...
+    `, [
+      retrievalId,
+      result.walletAddress,
+      result.success
+    ])
+  } catch (err) {
+    if (err.constraint === 'retrieval_results_retrieval_id_fkey') {
+      assert.fail(404, 'Retrieval Not Found')
+    } else if (err.constraint === 'retrieval_results_retrieval_id_key') {
+      assert.fail(409, 'Retrieval Already Completed')
```
juliangruber, line 59: hold on, this approach might not scale well enough :( Working on another PR, I just discovered a validation error where nothing useful is on the error object. So, I guess we mirror validations in JS?

juliangruber: see the update in https://github.com/filecoin-station/spark-api/pull/20

bajtos:
>the DB couldn't validate that either none or both have to be set.

I think this can be done using a trigger, but that's a lot of complexity. I prefer your solution more.

```sql
CREATE OR REPLACE FUNCTION validate_retrieval_result()
  RETURNS TRIGGER
  LANGUAGE plpgsql
  STABLE
  AS $function$
BEGIN
  IF (NEW.success IS NULL) THEN
    # validate new job with no result reported yet
    IF (NEW.finished_at IS NOT NULL) THEN
      RAISE EXCEPTION 'ERROR: cannot insert into column "finished_at" when "success" is not set.'
      USING DETAIL = 'You can set retrieval result fields only when "success" is set too.';
    END IF;
    # and so on for other fields
  ELSE
    # validate retrieval result
    IF (NEW.finished_at IS NULL) THEN
      RAISE EXCEPTION 'ERROR: "finished_at" is required when reporting retrieval result.';
    END IF;
    # and so on for other fields
  END IF;
END;
$function$;

CREATE TRIGGER on_retrievals_change_trigger
  BEFORE INSERT OR UPDATE ON retrievals
  FOR EACH ROW
  EXECUTE PROCEDURE validate_retrieval_result();
```

juliangruber:
Makes sense! I guess these conflicts are just part of the migration game

bajtos:
> Makes sense! I guess these conflicts are just part of the migration game

Yeah, that was my experience with DB migrations 🤷🏻 

In my last project, I prefixed migration ids with the last two digits of the current year, e.g. `23001`, to make filenames sortable lexicographically. Otherwise, we get files in the order `1.sql`, `10.sql`, `2.sql`, etc.

bajtos:
> In my last project, I prefixed migration ids with the last two digits of the current year, e.g. `23001`, to make filenames sortable lexicographically. Otherwise, we get files in the order `1.sql`, `10.sql`, `2.sql`, etc.

Ah, I see you are using three-digit ids like `006`, so all is good. Please disregard my last comment 🙈

bajtos (approved):
:shipit:

PR #20, opened 2023-05-31
Add retrieval result columns

juliangruber:


index.js
```diff
@@ -42,23 +41,38 @@ const setRetrievalResult = async (req, res, client, retrievalId) => {
...
+      new Date(result.endAt).getTime(),
+      result.byteLength
     ])
   } catch (err) {
     if (err.constraint === 'retrieval_results_retrieval_id_fkey') {
       assert.fail(404, 'Retrieval Not Found')
     } else if (err.constraint === 'retrieval_results_retrieval_id_key') {
       assert.fail(409, 'Retrieval Already Completed')
-    } else if (err.column) {
-      assert.fail(400, `Invalid .${toCamelCase(err.column)}`)
```
juliangruber, line 61: If for example you pass a non boolean `result.success`, there will be an error but it won't have a `err.column`. Therefore I replicated all type validations in JS

index.js
```diff
@@ -42,23 +41,38 @@ const setRetrievalResult = async (req, res, client, retrievalId) => {
...
+      )
     `, [
       retrievalId,
       result.walletAddress,
-      result.success
+      result.success,
+      new Date(result.startAt).getTime(),
+      result.statusCode,
+      new Date(result.firstByteAt).getTime(),
+      new Date(result.endAt).getTime(),
```
bajtos, line 68: I think you should validate that the date values are correctly formatted and return 400 error response if they are not.

Please include tests for this validation.

```
❯ node
Welcome to Node.js v18.14.2.
Type ".help" for more information.
> new Date('not-iso').getTime()
NaN
```

migrations/1685458879107.do.add-retrieval-result-columns.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE retrieval_results
+  ADD COLUMN start_at TIMESTAMP,
+  ADD COLUMN status_code INTEGER,
+  ADD COLUMN first_byte_at TIMESTAMP,
+  ADD COLUMN end_at TIMESTAMP,
```
bajtos, line 5: I think we should use `timestampz` to include the timezone information in the stored value.

See https://www.postgresql.org/docs/current/datatype-datetime.html

migrations/1685459848689.do.remove-timezones.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE retrieval_results
+  ALTER COLUMN start_at TYPE TIMESTAMP WITHOUT TIME ZONE,
+  ALTER COLUMN first_byte_at TYPE TIMESTAMP WITHOUT TIME ZONE,
+  ALTER COLUMN end_at TYPE TIMESTAMP WITHOUT TIME ZONE;
```
bajtos, line 4: Can we merge these two migrations into a single one, please, so that we have only one migration per PR?

test/test.js
```diff
@@ -78,28 +78,56 @@ describe('Routes', () => {
...
       `, [
         retrievalId
       ])
-      assert.strictEqual(updatedRetrievalRow.success, true)
-      assert.strictEqual(updatedRetrievalRow.wallet_address, walletAddress)
+      assert.strictEqual(retrievalResultRow.success, result.success)
+      assert.strictEqual(retrievalResultRow.wallet_address, walletAddress)
+      assert.strictEqual(
+        Math.round(new Date(retrievalResultRow.start_at).getTime() / 1000),
+        result.startAt.getTime()
```
bajtos, line 110: I am surprised that we are not getting Date instances from the database. Why is that? Is it because we are storing the date-time values without any timezone info?

bajtos (changes requested):

test/test.js
juliangruber: I was doing a bunch of mistakes while trying to get the tests to pass. Thanks for bringing it up, we should be handling timestamps correctly now

bajtos: 👌🏻

test/test.js
```diff
@@ -78,28 +78,56 @@ describe('Routes', () => {
...
+      assert.strictEqual(retrievalResultRow.wallet_address, walletAddress)
+      assert.strictEqual(
+        retrievalResultRow.start_at.getTime(),
+        result.startAt.getTime()
+      )
+      assert.strictEqual(retrievalResultRow.status_code, result.statusCode)
+      assert.strictEqual(
+        retrievalResultRow.first_byte_at.getTime(),
+        result.firstByteAt.getTime()
+      )
```
bajtos, line 116: What are the benefits of comparing `getTime` values instead of the actual `Date` values? If we get different values and the test fails, it will be difficult to understand what's different when the error message shows milliseconds since Epoch. (For example, when the time is one hour off because of timezone problems.) Have you considered using `toISOString()` or `toJSON` instead of `getTime()`?

```suggestion
      assert.strictEqual(
        retrievalResultRow.first_byte_at.toISOString(),
        result.firstByteAt.toISOString()
      )
```

This is a nitpick; feel free to ignore it.

bajtos (approved):
👍🏻

migrations/1685529110405.do.add-retrieval-result-columns.sql
bajtos, line 1: I think you should change this filename to something like `007.do.add-retrieval-result-columns.sql`, per our discussion in the other PR.

bajtos: Also there is no `undo` migration file. Maybe you should add it too?

test/test.js
juliangruber: > What are the benefits of comparing getTime values instead of the actual Date values?

It is because strict equality checks for two `Date` objects fail even if they point at the same time, since they're different identities:

```js
> const d = new Date()
undefined
> const d2 = new Date(d.getTime())
undefined
> d === d2
false
> d
2023-05-31T13:44:34.313Z
> d2
2023-05-31T13:44:34.313Z
> 

```

We can use iso string comparison too, yeah that should be more readable

migrations/1685529110405.do.add-retrieval-result-columns.sql
juliangruber: I decided not to add `undo` any more for now since it's work and we don't have any data yet for which we need the step back

PR #22, opened 2023-05-31
test: remove log output

juliangruber:
Remove logging from test output

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
   },
   "devDependencies": {
     "mocha": "^10.2.0",
+    "noop-logger": "^0.1.1",
     "standard": "^17.1.0",
     "varint": "^6.0.0"
   },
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -12,6 +12,7 @@
   },
   "devDependencies": {
     "mocha": "^10.2.0",
+    "noop-logger": "^0.1.1",
```
bajtos, line 15: Since we need only `.info` and `.error`, I'd prefer write the noop-logger manually instead of adding a new dependency. Also considering that noop-logger haven't been updated in 7 years and the GH repository was archived on Jun 2, 2021.

bajtos (approved):
Having a clean test output is a good thing, thank you for caring!

package.json
juliangruber: I looked at the repo too, and thought just the opposite :D To me it's fear to archive it, because the project is one of those few that can be considered done. Using a package over custom implementation also has the advantage that it gives the concept a name "ah let's use a noop logger", and a README with some extra context.

If you feel strongly about this I'll inline the logger, but maybe my reasoning changed that?

PR #25, opened 2023-06-14
add provider peer id

juliangruber:
Lassie requires the peer id in addition to the provider address, see https://github.com/filecoin-project/lassie/blob/main/docs/HTTP_SPEC.md#providers-request-query-parameter.


index.js
```diff
@@ -19,9 +19,10 @@ const handler = async (req, res, client) => {
 const createRetrieval = async (res, client) => {
   // TODO: Consolidate to one query
   const { rows: [retrievalTemplate] } = await client.query(`
-    SELECT id, cid, provider_address, protocol
+    SELECT id, cid, provider_address, peer_id, protocol
     FROM retrieval_templates
-    OFFSET floor(random() * (SELECT COUNT(*) FROM retrieval_templates))
+    WHERE enabled = TRUE
```
juliangruber, line 24: I couldn't find a peer id for all retrieval templates. I didn't want to delete the rows either though, otherwise we'd have `retrieval_results` without a matching `retrieval_template`. Or shall we just delete the retrieval results?

README.md
```diff
@@ -16,6 +16,7 @@ Response:
   id: String,
   cid: String,
   providerAddress: String,
+  peerID: String,
```
bajtos, line 19: Let's call this `providerPeerID` to make it less ambiguous and more consistent with `providerAddress`. 

Other SPARK components may get their peer IDs in the future, too. At that point, it will become difficult to tell which peer is the `peerID` belonging to.

Also: have you considered encoding the peerID into the `providerAddress`? Both options have different pros and cons. I am OK with adding another field, as you are doing now.

index.js
bajtos: I think this depends on whether we want to allow retrievals from providers with unknown peer IDs going forward. If we want to make peer id required, then I think it's ok to delete old retrieval results.

One thing to consider to make it easier for us to evolve the database in the future: we can denormalise our schema, copy the relevant `retrieval_template` fields into each `retrieval_results` row, and remove the relation between those tables.

bajtos: Having said that, I think we will need to allow removing records from our templates anyway, i.e. to let future jobs choose from a different set of available templates. The flag `enabled` solves this requirement too.

If we change the flag to `deleted: true`, then we get the standard soft-delete design.

bajtos (approved):

README.md
bajtos: > Other SPARK components may get their peer IDs in the future, too. At that point, it will become difficult to tell which peer is the peerID belonging to.


E.g. each Station module instance already has a peer ID used for libp2p connections. See [`Zinnia.peerId`](https://github.com/filecoin-station/zinnia/blob/main/docs/building-modules.md#zinniapeerid)

index.js
juliangruber: > I think this depends on whether we want to allow retrievals from providers with unknown peer IDs going forward. If we want to make peer id required, then I think it's ok to delete old retrieval results.

We have to do this yes, as Lassie requires the peer ID in addition to its address

juliangruber: Sg, I'll change `enabled` to `deleted`, exactly for that forward compatibility design

bajtos: > Having said that, I think we will need to allow removing records from our templates anyway, i.e. to let future jobs choose from a different set of available templates. The flag enabled solves this requirement too.

I was trying to say that adding a new flag like `enabled` or `deleted` is a good idea, as it will be helpful for other things too.

bajtos (approved):
:shipit:

PR #56, opened 2023-08-30
feat: add round & IP address to data model

bajtos:
Add two new columns to `retrievals` table:
```
created_at_round BIGINT NOT NULL
created_from_address INET NOT NULL
```

Add two new columns to `retrieval_results` table:
```
completed_at_round BIGINT NOT NULL
completed_from_address INET NOT NULL
```

Rework the HTTP handler and the test suite to fill these new columns, using a hard-coded round number (for now).

For the existing rows, use round `0` and address `0.0.0.0`.

_This is the first step towards https://github.com/filecoin-station/spark/issues/13._

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,12 +4,12 @@ import getRawBody from 'raw-body'
 import assert from 'http-assert'
 import { validate } from './lib/validate.js'
 
-const handler = async (req, res, client) => {
+const handler = async (req, res, client, getCurrentRound) => {
   const segs = req.url.split('/').filter(Boolean)
   if (segs[0] === 'retrievals' && req.method === 'POST') {
-    await createRetrieval(req, res, client)
+    await createRetrieval(req, res, client, getCurrentRound)
   } else if (segs[0] === 'retrievals' && req.method === 'PATCH') {
-    await setRetrievalResult(req, res, client, Number(segs[1]))
+    await setRetrievalResult(req, res, client, Number(segs[1]), getCurrentRound)
   } else if (segs[0] === 'retrievals' && req.method === 'GET') {
     await getRetrieval(req, res, client, Number(segs[1]))
   } else {
@@ -17,7 +17,8 @@ const handler = async (req, res, client) => {
   }
 }
 
-const createRetrieval = async (req, res, client) => {
+const createRetrieval = async (req, res, client, getCurrentRound) => {
+  const round = await getCurrentRound()
   const body = await getRawBody(req, { limit: '100kb' })
   const meta = body.length > 0 ? JSON.parse(body) : {}
   validate(meta, 'sparkVersion', { type: 'string', required: false })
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -36,14 +37,18 @@ const createRetrieval = async (req, res, client) => {
...
-    VALUES ($1, $2, $3)
+    VALUES ($1, $2, $3, $4, $5)
     RETURNING id
   `, [
     retrievalTemplate.id,
     meta.sparkVersion,
-    meta.zinniaVersion
+    meta.zinniaVersion,
+    round,
+    req.connection.remoteAddress
```
bajtos, line 51: What are the implication for user's privacy when we start collecting their IP addresses?

migrations/012.do.round-and-ip-address.sql
```diff
@@ -0,0 +1,15 @@
...
+-- Postgres and SQL do not support unsigned integers
+ALTER TABLE retrievals ADD COLUMN created_at_round BIGINT;
+ALTER TABLE retrievals ADD COLUMN created_from_address INET;
+
+UPDATE retrievals SET created_at_round = 0, created_from_address = '0.0.0.0';
+ALTER TABLE retrievals ALTER COLUMN created_at_round SET NOT NULL;
+ALTER TABLE retrievals ALTER COLUMN created_from_address SET NOT NULL;
+
+ALTER TABLE retrieval_results ADD COLUMN completed_at_round BIGINT;
+ALTER TABLE retrieval_results ADD COLUMN completed_from_address INET;
```
bajtos, line 11: I am not sure yet whether we need to collect both "created" and "completed" round & address and how to handle the case when they are in conflict.

To make it easier to observe the real network of Stations, I prefer to collect them both, inspect the data and decide on the next steps based on that.

index.js
juliangruber: looks like we need to add a privacy policy https://termly.io/faq/is-ip-address-personal-information-under-ccpa/#:~:text=IP%20addresses%20are%20classified%20as,this%20in%20a%20privacy%20policy., cc @coreymjames

migrations/012.do.round-and-ip-address.sql
juliangruber: I was wondering the same and came to the same conclusion 👍

juliangruber (approved):

juliangruber (changes requested):
actually, requesting changes, as we need to ensure we have a privacy policy (I think) before merging this

bajtos:
> actually, requesting changes, as we need to ensure we have a privacy policy (I think) before merging this

I had a gut feeling it's not a good idea to add both rounds and IP addresses together in one PR. Now I see why 😩

I am going to split this PR into two, so that we can land round numbers sooner.

bajtos:
Closing in favour of #58 and #59

PR #58, opened 2023-08-31
feat: add round number to the data model

bajtos:
Add a new column to `retrievals` table:
```
created_at_round BIGINT NOT NULL
```

Add a new column to `retrieval_results` table:
```
completed_at_round BIGINT NOT NULL
```

Rework the HTTP handler and the test suite to fill these new columns,
using a hard-coded round number (for now).

_This is the first step towards filecoin-station/spark#13. It replaces #56._

PR #59, opened 2023-08-31
feat: add IP address to the data model

bajtos:
Add a new column to `retrievals` table:
```
created_from_address INET NOT NULL
```

Add a new column to `retrieval_results` table:
```
completed_from_address INET NOT NULL
```

Rework the HTTP handler and the test suite to fill these new columns.

_This PR replaces #56, see also #58._

### DIFF
--- a/index.js
+++ b/index.js
@@ -48,7 +48,7 @@ const createRetrieval = async (req, res, client, getCurrentRound) => {
     meta.sparkVersion,
     meta.zinniaVersion,
     round,
-    req.connection.remoteAddress
+    getClientAddress(req)
   ])
   json(res, {
     id: retrieval.id,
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -36,14 +36,18 @@ const createRetrieval = async (req, res, client) => {
...
-    VALUES ($1, $2, $3)
+    VALUES ($1, $2, $3, $4, $5)
     RETURNING id
   `, [
     retrievalTemplate.id,
     meta.sparkVersion,
-    meta.zinniaVersion
+    meta.zinniaVersion,
+    round,
+    req.connection.remoteAddress
```
bajtos, line 50: _Re-posting https://github.com/filecoin-station/spark-api/pull/56#discussion_r1310404604_

> > What are the implications for the privacy of Station's users when we start collecting their IP addresses?
> 
> looks like we need to add a privacy policy https://termly.io/faq/is-ip-address-personal-information-under-ccpa/#:~:text=IP%20addresses%20are%20classified%20as,this%20in%20a%20privacy%20policy., cc @coreymjames

bajtos: Would anything change if we collect only the IPv4 /24 subnet instead of the full IPv4 address? Is the subnet considered as a personal information too?

juliangruber: Good question! @coreymjames could you please help us figure out the legal implications?

PR #61, opened 2023-08-31
feat: integrate with MERidian RoundStarted events

bajtos:
- feat: map MERidian rounds to SPARK rounds
- feat: listen for MERidian contract events

_This work is part of https://github.com/filecoin-station/spark/issues/13_


migrations/013.do.meridian-rounds.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE meridian_rounds (
+  -- 0x of f4 address of the MERidian smart contract emitting "round started" events
+  -- We are using TEXT instead of BYTEA for simplicity
+  contract_address TEXT NOT NULL PRIMARY KEY,
```
juliangruber, line 4: it looks like this table is storing contract version rather than rounds. With meridian rounds I'd expect the primary key to be a composite of contract and round offset. Like this we will only store one row per contract (right?)

juliangruber: I'm somewhat confused about this logic (and therefore round-tracker). Could you please give me a rough overview of the algorithm, so that I can continue with the review?

juliangruber (changes requested):

migrations/013.do.meridian-rounds.sql
bajtos: Maybe it's better to start the review in the function using this data: https://github.com/filecoin-station/spark-api/pull/61/files#diff-16e2f7b1cd688374798168c1ef69985dbd460a6a451142a27b29c5d8c348312dR22

What's the best place to write down documentation for this logic? Should I create a new ADR?

---

There are three cases I want to handle:

1. Business as usual - the IE contract advanced the round by one
2. Fresh start, e.g. a new spark-api instance is deployed, or we deploy this PR to an existing instance.
3. Upgrade of the IE contract

For each IE version (defined as the smart contract address), I am keeping track of three fields:
- `contractAddress` 
- `sparkRoundOffset`
- `lastSparkRound`

Whenever a new IE round is started, we know the current IE round number (`meridianRoundNumber`)

Let me explain how are the different cases handled.

**Business as usual**

We want to map IE round number to SPARK round number. This assumes we have already initialised our DB for the current IE contract version we are working with.

```
sparkRound = meridianRoundNumber + sparkRoundOffset
```

For example, if we deploy my PR and then observe IE round 123, then `sparkRoundOffset` is `-122` and we calculate the spark round as
```
123 + (-122) = 1
```

We update the record for the current IE contract address to set `last_spark_round = sparkRound`.

**Fresh start**

There is no record in our DB. We want to map the current IE round number to SPARK round 1. Also, we want to setup `sparkRoundOffset` so that the algorithm above produces correct SPARK round numbers.

```
sparkRound = 1
sparkRoundOffset = sparkRound - meridianRoundNumber
```

We insert a new record to our DB with the address of the current IE contract, `sparkRoundOffset`, and `last_spark_round = sparkRound`.

**Upgrading IE contract**

We have one or more existing records in our DB. We know what is the last SPARK round that we calculated from the previous version of the IE contract (`lastSparkRound`). We also know what is the round number of the new IE contract.

```
sparkRound = lastSparkRound + 1
sparkRoundOffset = sparkRound - meridianRoundNumber
```

We insert a new record to our DB with the address of the current IE contract, `sparkRoundOffset`, and `last_spark_round = sparkRound`.

If you are wondering how to find out what is the last SPARK round that we calculated from the previous version of the IE contract - we can easily find it in our DB:

```sql
SELECT last_spark_round FROM meridian_rounds ORDER BY last_spark_round DESC LIMIT 1
```

lib/round-tracker.js
```diff
@@ -0,0 +1,78 @@
...
+  const updateSparkRound = async (meridianRound) => {
+    sparkRound = await mapCurrentMeridianRoundToSparkRound({
+      meridianContractAddress: contract.address,
+      meridianRound: BigInt(meridianRound),
+      pgClient
+    })
+    console.log('SPARK round started: %s', sparkRound)
+  }
+
+  contract.on('RoundStart', updateSparkRound)
```
juliangruber, line 16: `updateSparkRound` is an `async` function and I'm not aware of `ethers` having some logic to properly handle async event handlers. Therefore I think this needs a `.catch`

lib/round-tracker.js
```diff
@@ -0,0 +1,78 @@
...
+
+export async function mapCurrentMeridianRoundToSparkRound ({
+  meridianContractAddress,
+  meridianRound,
+  pgClient
+}) {
+  let sparkRound
+
+  const { rows: [lastMeridianRecord] } = await pgClient.query(
+    'SELECT * FROM meridian_contract_versions ORDER BY last_spark_round DESC LIMIT 1'
```
juliangruber, line 30: We could also query the new `spark_rounds` table for this, does any option have an advantage?

lib/round-tracker.js
```diff
@@ -0,0 +1,78 @@
...
+      [sparkRound, meridianContractAddress]
+    )
+    console.log('Mapped %s IE round %s to SPARK round %s',
+      meridianContractAddress,
+      meridianRound,
+      sparkRound
+    )
+  } else {
+  // We are running for the first time and need to map the meridian round to spark round 1
+  // Or the contract address has changed
```
bajtos, line 47: 😠 I expected that `standard --fix` would indent this correctly.

lib/round-tracker.js
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export async function mapCurrentMeridianRoundToSparkRound ({
+  meridianContractAddress,
+  meridianRound,
+  pgClient
+}) {
+  let sparkRound
+
+  const { rows: [lastMeridianRecord] } = await pgClient.query(
```
juliangruber, line 29:
```suggestion
  const { rows: [lastSparkRoundContract] } = await pgClient.query(
```

I think this fits the new names better

lib/round-tracker.js
```diff
@@ -0,0 +1,78 @@
...
+    'SELECT * FROM meridian_contract_versions ORDER BY last_spark_round DESC LIMIT 1'
+  )
+
+  // More events coming from the same meridian contract
+  if (lastMeridianRecord && lastMeridianRecord.contract_address === meridianContractAddress) {
```
juliangruber, line 34:
```suggestion
  if (lastMeridianRecord?.contract_address === meridianContractAddress) {
```

lib/round-tracker.js
```diff
@@ -0,0 +1,78 @@
...
+    sparkRound = BigInt(lastMeridianRecord.spark_round_offset) + meridianRound
```
juliangruber, line 35:
```suggestion
    sparkRoundIndex = BigInt(lastMeridianRecord.spark_round_offset) + meridianRound
```

What do you think about making clear we are talking about the number? Could be id, index, number?

lib/round-tracker.js
juliangruber, line 47:
```suggestion
    // We are running for the first time and need to map the meridian round to spark round 1
    // Or the contract address has changed
```

lib/round-tracker.js
```diff
@@ -0,0 +1,78 @@
...
+    console.log('Mapped %s IE round %s to SPARK round %s',
+      meridianContractAddress,
+      meridianRound,
+      sparkRound
+    )
+  } else {
+  // We are running for the first time and need to map the meridian round to spark round 1
+  // Or the contract address has changed
+    const lastSparkRound = BigInt(lastMeridianRecord?.last_spark_round ?? 0)
+    sparkRound = lastSparkRound + 1n
```
juliangruber, line 49: awesome that you're using the native BigInt type 👏  I'm going to refactor the code I wrote to use this too

juliangruber (changes requested):

lib/round-tracker.js
bajtos: Great catch! (No pun intended 🤣)

lib/round-tracker.js
bajtos: We need to find the `contract_address` we used for the previous SPARK round. That information is not stored in the `spark_rounds` table.

lib/round-tracker.js
bajtos: I see. Let me improve this a bit more to `contractVersionOfPreviousSparkRound`.

lib/round-tracker.js
bajtos: I am okay with that. Let's use "number" because SPARK rounds start from the number 1. The value `0` is reserved for existing measurements not connected to any round yet.

lib/round-tracker.js
bajtos: Cool!

FWIW, Ethers v6 migrated to `BigInt`, it's a bummer we have to use v5 for now.

https://docs.ethers.org/v6/migrating/#migrate-bigint

migrations/013.do.meridian-rounds.sql
bajtos: I captured the explanation in a code comment.

lib/round-tracker.js
juliangruber: Yes I think so

lib/round-tracker.js
```diff
@@ -0,0 +1,154 @@
...
+We insert a new record to our DB with the address of the current IE contract, `sparkRoundOffset`,
+and `last_spark_round = sparkRoundNumber`.
+
+If you are wondering how to find out what is the last SPARK round that we calculated from the
+previous version of the IE contract - we can easily find it in our DB:
+
+```sql
+SELECT last_spark_round FROM meridian_contract_versions ORDER BY last_spark_round DESC LIMIT 1
+```
+*/
```
juliangruber, line 92: 👏

lib/round-tracker.js
```diff
@@ -0,0 +1,154 @@
...
+    console.log('Mapped %s IE round %s to SPARK round %s',
+      meridianContractAddress,
+      meridianRoundIndex,
+      sparkRoundNumber
+    )
+  } else {
+    // We are running for the first time and need to map the meridian round to spark round 1
+    // Or the contract address has changed
+    const lastSparkRound = BigInt(contractVersionOfPreviousSparkRound?.last_spark_round ?? 0)
+    sparkRoundNumber = lastSparkRound + 1n
```
juliangruber, line 121:
```suggestion
    const lastSparkRoundNumber = BigInt(contractVersionOfPreviousSparkRound?.last_spark_round ?? 0)
    sparkRoundNumber = lastSparkRoundNumber + 1n
```

juliangruber (changes requested):

juliangruber (approved):

PR #63, opened 2023-09-01
Add publish service. Closes #42

### LINKED ISSUE #42: Meridian: Measure commit phase
```[tasklist]
### Tasks
- [x] Aggregate measurements
- [x] Expose via IPFS
- [x] Send CID to IE contract
- [ ] Await round rewards
- [ ] Clean up aggregates
```

juliangruber:
Closes #42

@bajtos happy to fulfil change requests or hand over to you


.github/workflows/ci.yml
```diff
@@ -27,13 +26,50 @@ jobs:
...
+        with:
+          node-version: 20
+      - run: npm ci
+      - run: npm run migrate
+      - run: cd spark-publish && npm ci
+      - run: cd spark-publish && npm test
+  deploy-api:
     if: github.ref == 'refs/heads/main'
-    needs: build
+    needs: [build-api, build-publish]
```
bajtos, line 58: IIUC, this config allows deployments of new spark-api versions only if both spark-api and spark-publish pass the tests. (And vice versa for deploying new spark-publish versions.)

The other option is to allow spark-api deployments as long as spark-api tests are passing, and similarly for spark-publish deployments.

I am fine either way; just wanted to make sure this is a conscious decision on our side.

migrations/013.do.published-as.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE retrievals ADD COLUMN published_as VARCHAR(64);
```
bajtos, line 1: Can we rely on CIDs to be at most 64 characters long?

According to https://maximorlov.com/char-varchar-text-postgresql/, which quotes from the official Postgres docs, there is no performance benefit in using `char` or `varchar` instead of `text`.

> There is no performance difference among these three types, apart from increased storage space when using the blank-padded type, and a few extra CPU cycles to check the length when storing into a length-constrained column. While character(n) has performance advantages in some other database systems, there is no such advantage in PostgreSQL; in fact character(n) is usually the slowest of the three because of its additional storage costs. In most situations text or character varying should be used instead.

```suggestion
ALTER TABLE retrievals ADD COLUMN published_as TEXT;
```

package.json
```diff
@@ -7,6 +7,7 @@
   "type": "module",
   "description": "API for SPARK",
   "scripts": {
+    "migrate": "node bin/migrate.js",
```
bajtos, line 10: 👏🏻

spark-publish/abi.json
bajtos, line 1: I will need this file for my work on https://github.com/filecoin-station/spark-api/pull/61, too.

I am proposing to extract the bits we need for interaction with the IE contract into a component shared by both `spark-api` and `spark-publish`.

WDYT?

A good argument against my proposal is if we want the spark-api and spark-measurement services to talk to a different version of the IE contract.

spark-publish/test/test.js
```diff
@@ -0,0 +1,124 @@
...
+    const cid = 'bafybeicmyzlxgqeg5lgjgnzducj37s7bxhxk6vywqtuym2vhqzxjtymqvm'
+
+    const client = {
+      async query (_, params) {
+        assert.strictEqual(params[0], 1)
+
+        client.query = (_, params) => {
+          assert.strictEqual(params[0], cid)
+          assert.deepStrictEqual(params[1], [])
+        }
```
bajtos, line 19: I think we discussed this in the past. I have a bad experience with assertions inside mocked/stubbed functions. If the function is never called, then we never assert anything and the test passes, giving us a false sense of security.

A relatively easy solution:

```js
const dbQueryParams = []
client.query = (_, params) => { dbQueryParams.push(params) })
// ...

await publish({/*...*/}

assert.deepEquals(dbQueryParams, [
  [cid, []],
])
```

spark-publish/test/test.js
```diff
@@ -0,0 +1,124 @@
...
+
+        return { rows: [] }
+      }
+    }
+
+    const web3Storage = {
+      async put (files) {
+        assert.strictEqual(files.length, 1)
```
bajtos, line 27: Ditto.

spark-publish/test/test.js
```diff
@@ -0,0 +1,124 @@
...
+        return CID.parse(cid)
+      }
+    }
+
+    const ieContract = {
+      async addMeasurements (_cid) {
+        assert.strictEqual(_cid, cid)
```
bajtos, line 34: Ditto.

spark-publish/test/test.js
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    await publish({
+      client,
+      web3Storage,
+      ieContract,
+      maxMeasurements: 1,
+      logger
+    })
+  })
```
bajtos, line 123: IMO, the integration test should inspect data committed on the chain (e.g. the local chain run by Anvil).

Can you please add a TODO comment here?

bajtos (commented):
Nice work, @juliangruber 👏🏻 

I have a few comments to discuss, PTAL 👇🏻

migrations/013.do.published-as.sql
juliangruber: I'm not very certain how large a CID can be. This is definitely safer, so committed it 👍 

Space is a consideration here, since our spark db instance sometimes runs out of memory. I don't think we have to be this precise though

.github/workflows/ci.yml
juliangruber: The primary reason for having both projects in one repo is that they will always be deployed together, that's why I set it up like this. If we are fine with deploying them separately, I'd move them to different repos. Wdyt?

spark-publish/abi.json
juliangruber: If it's just this one file, then I think copy pasta is fine. Otherwise yeah I thought about this too, a `spark-contract` repo could make sense. It makes us update more place though when something re the contract changes.

spark-publish/test/test.js
juliangruber: good point, I was hoping to get your input on this

bajtos (approved):
👏🏻 :shipit:

bajtos:
@juliangruber I think you need to update branch protection rules, remove `build` and add the two new `build-*` steps instead.

juliangruber:
good point, will do!

juliangruber:
```bash
$ cd spark-publish
$ fly logs
[...]
2023-09-04T12:53:18Z app[6e82994bee5798] cdg [info]Publishing 1000 measurements
2023-09-04T12:53:21Z app[6e82994bee5798] cdg [info]Measurements packaged in bafybeifhxc22tanwfa7yi5b6w24tgzynoqldsdqx6yowudsxkn3pwpf2ju
2023-09-04T12:53:21Z app[6e82994bee5798] cdg [info]ie.addMeasurements()...
2023-09-04T12:54:02Z app[6e82994bee5798] cdg [info]Done!
```

juliangruber:
🥳

spark-publish/abi.json
bajtos: I was thinking about a folder in this repo, shared by both components. E.g. `spark-api/meridian-contract`. Anyhow, I'll take a look at this while working on #61.

PR #71, opened 2023-09-06
refactor: share Meridian config

bajtos:
Rework spark-api to use the contract configuration from spark-publish.


spark-publish/ie-contract-config.js
juliangruber, line 1: I don't think `spark-publish` should own the contract. It is a sub project of `spark-api`, and both need access to the contract. Therefore I think the contract should be either owned by `spark-api` or none (a separate project)

juliangruber (changes requested):

spark-publish/ie-contract-config.js
bajtos: I see your point. Moving the contract to a different place requires complex changes, because spark-publish's Dockerfile is rooted inside spark-publish directory and does not have access to other files in this repo.

The way I am looking at this:
- spark-publish is interacting with the contract (submitting measurements), therefore it makes sense to me that spark-publish owns the contract config
- spark-api wants to listen for events created by the contract used by spark-publish, therefore it's ok to load contract config from spark-publish

Anyhow, I am fine to move the contract config to a new directory shared by both spark-api and spark-publish, if you think it's worth the extra effort.

juliangruber: Ok, let's keep it like this, knowing we both want to refactor in the future, once the need arises

juliangruber (approved):

PR #72, opened 2023-09-06
fix: forward RoundStarted event args

bajtos:
Fix the bug where we were not able to handle SparkRound events.


juliangruber (commented):
👍

PR #76, opened 2023-09-07
bump MAX_MEASUREMENTS_PER_ROUND to 2000

juliangruber:
We're comfortably processing batches of 1000, let's bump to 2000. See https://fly.io/apps/spark-publish/metrics for metrics

### DIFF
--- a/spark-publish/Dockerfile
+++ b/spark-publish/Dockerfile
@@ -12,6 +12,7 @@ WORKDIR /app
 # Set production environment
 ENV NODE_ENV="production"
 ENV SENTRY_ENVIRONMMENT="production"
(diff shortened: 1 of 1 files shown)


juliangruber:
should be fine without review

bajtos:
The change LGTM 👍🏻 It was fine to land it without review.

FWIW, in [spark-publish logs](https://fly.io/apps/spark-publish/monitoring), I see that we typically send 15-25 measurements in a single batch. I think this PR is not going to change anything 🤷🏻‍♂️

juliangruber:
Ah, now that it has caught up yes, but this was super useful for when we had millions of uncommitted rows still :)

juliangruber:
And will be useful once we have a bazillion clients

PR #80, opened 2023-09-11
feat: round-based tasking

bajtos:
Add a new DB table and HTTP API endpoints to implement the new tasking design
based on MERidian rounds.

1. Whenever a new MERidian round triggers the start of a new SPARK round,
   we define a list of tasks for this round. This list is fixed for the entire
   duration of the round.

2. SPARK checkers will call the new HTTP API endpoint to obtain details about
   the current round and the tasks scheduled.

   ```
   GET /rounds/current
   ```

3. The Evaluate service will call the new HTTP API endpoint to obtain details
   about the evaluated round (e.g. 12345) for fraud-detection purposes.

   ```
   GET /rounds/12345
   ```

The response provides:
 - round number (important when requesting the current round)
 - time when the round should end
 - list of tasks (`{cid, provider_address, protocol}`) for this round

Connect to https://github.com/filecoin-station/spark/issues/13


bajtos:
This PR does not integrate existing SPARK clients - they keep calling the old `POST /measurements` endpoint. If we don't do any integration, then these measurements will be eventually flagged as fraudulent by our evaluation service. I think that's actually a good thing, as it allows us to test our evaluation logic both for happy and fraudulent paths.

Either way, I will need both existing SPARK clients and the future new version of SPARK to write measurements to the same SQL table, and this SQL table needs to be decoupled from `retrievals` and `retrievals_templates`. I'll look into this in a standalone PR.

index.js
```diff
@@ -4,6 +4,9 @@ import getRawBody from 'raw-body'
 import assert from 'http-assert'
 import { validate } from './lib/validate.js'
 
+// Approximate length of each SPARK/Meridian round in milliseconds
+const ROUND_LENGTH_IN_MS = 30_000 /* ms */
```
bajtos, line 8: @juliangruber Is there a way to read this information from the contract? Is it the `nextRoundLength` variable?

Do you think it's worth the effort to remove this constant and query the round length whenever we detect the start of a new MERidian round?

Considering how infrequently we are going to change this constant, maybe it's fine to keep this hard-coded for now.

bajtos: BTW, the smart contract code says that the next round length is 10:

https://github.com/Meridian-IE/impact-evaluator/blob/9242c15bfeafcc89a23e91caa34532b4bd09850a/src/ImpactEvaluator.sol#L17
```sol
    uint public nextRoundLength = 10;
```

juliangruber: > @juliangruber Is there a way to read this information from the contract? Is it the nextRoundLength variable?

Yes! Any public variable has a getter automatically. You can get it via `ie.nextRoundLength()`

> Do you think it's worth the effort to remove this constant and query the round length whenever we detect the start of a new MERidian round?
Considering how infrequently we are going to change this constant, maybe it's fine to keep this hard-coded for now.

We could either query at the start of each round, or at process start? It should be fine to query at the start of each round, since the call is free and should be reasonably fast.

> BTW, the smart contract code says that the next round length is 10 (~5min?)

Yes, this is the default state. I've since changed the round length via `ie.setNextRoundLength(uint _nextRoundLength)`

index.js
```diff
@@ -4,6 +4,9 @@ import getRawBody from 'raw-body'
 import assert from 'http-assert'
 import { validate } from './lib/validate.js'
 
+// Approximate length of each SPARK/Meridian round in milliseconds
+const ROUND_LENGTH_IN_MS = 30 * 60_000 /* ms */
```
juliangruber, line 8: as discussed, this can be queried from the contract:

```suggestion
const ROUND_LENGTH_IN_MS = (await ie.nextRoundLength()) * 30_000 /* block time ms */
```

index.js
```diff
@@ -161,6 +166,47 @@ const getRetrieval = async (req, res, client, retrievalId) => {
...
+  if (!round) {
+    return notFound(res)
+  }
+
+  const { rows: tasks } = await client.query('SELECT * FROM tasks WHERE round_id = $1', [roundNumber])
+
+  json(res, {
+    roundId: roundNumber.toString(),
+    endsBefore: new Date(round.created_at.getTime() + ROUND_LENGTH_IN_MS),
+    tasks: tasks.map(it => ({
```
juliangruber, line 186: what does `it` stand for here?

lib/round-tracker.js
```diff
@@ -1,6 +1,8 @@
 import Sentry from '@sentry/node'
 import { createMeridianContract } from './meridian/contract.js'
 
+export const TASKS_PER_ROUND = 30
```
juliangruber, line 4: Is there a relationship between this variable and the round length?

migrations/015.do.tasks-per-round.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE tasks (
```
juliangruber, line 1: `tasks` is quite generic, what about `retrieval_tasks`?

juliangruber (changes requested):

index.js
bajtos: It stands for whatever we are iterating over - the element we are mapping.

I was using `t`, but then I realised it's potentially confusing because we use `t` for the current test instance in TAP-like tests too.

I also don't want to use `task` because it's too similar to `tasks`.

What name do you use in cases like this?

lib/round-tracker.js
bajtos: Sort of. We want to choose a number that will, on the one hand, allow most Stations to complete all tasks within one round and, on the other hand, keep Stations busy most of the time.

See https://www.notion.so/pl-strflt/SPARK-tasking-v2-604e26d57f6b4892946525bcb3a77104?pvs=4#ded1cd98c2664a2289453d38e2715643 for more details, this constant represents `TC` (tasks per committee).

We will need to tweak this value based on measurements; that's why I put it here as a constant.

I think this value will need to be much, much higher, like the low hundreds. However, we can't do that now because we have only 200 retrieval templates to sample from.

index.js
bajtos: > Yes, this is the default state. I've since changed the round length via `ie.setNextRoundLength(uint _nextRoundLength)`

Yeah, I remember. It makes me wonder though, if we will remember to bump it up whenever we deploy a new version of the smart contract 🤔

index.js
juliangruber: Interesting! To me `it` means iterable/iterator. See for example https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols?retiredLocale=de#simple_iterator

In this case no one is right of course. I think I would have used `task` or `t`. I don't feel strongly about this

lib/round-tracker.js
juliangruber: got it, thank you! Could you add this context as a comment?

index.js
juliangruber: I wondered about this too. A spark deployment script would be great, which is independent of the meridian deployment script

juliangruber: However, the manual process works for me still, since
1. We expect to deploy the contract only once
2. It's not hard to course correct

index.js
bajtos: > > Do you think it's worth the effort to remove this constant and query the round length whenever we detect the start of a new MERidian round?
Considering how infrequently we are going to change this constant, maybe it's fine to keep this hard-coded for now.
>
> We could either query at the start of each round, or at process start? It should be fine to query at the start of each round, since the call is free and should be reasonably fast.

I am not very comfortable querying this only once at the process start because contract admins can dynamically change the value.

I am going to query this whenever a new round starts.

bajtos: See [1cbf01d](https://github.com/filecoin-station/spark-api/pull/80/commits/1cbf01d7a1aa92b9d608848849b3f7a74347a65a)

lib/round-tracker.js
```diff
@@ -146,14 +154,55 @@ export async function mapCurrentMeridianRoundToSparkRound ({
     )
   }
 
-  await pgClient.query(`
+  // TODO: replace "new Date()" with the time when the NextRound event was committed on the chain
```
juliangruber, line 157: To my knowledge there's no reliable time on chain

juliangruber (changes requested):

index.js
bajtos: Renamed all `it` variables to `t` in 8b925b3

lib/round-tracker.js
bajtos: Added a comment to explain - see [390249c](https://github.com/filecoin-station/spark-api/pull/80/commits/390249cb1ea33284568f427f704920ef3bd55be0)

lib/round-tracker.js
bajtos: > To my knowledge there's no reliable time on chain

I know.

Stations are also not able to stop the work exactly at the round end because they don't know how long the next retrieval check is going to take. 

I consider the round deadline an approximate value, a hint to Stations.

An alternative we can consider:
- Whenever a Station wants to start the next retrieval check, it calls SPARK API to obtain the current round (and tasks for that round). If the round number is the same as before, then it continues in selecting samples for that same round. If the round number goes up, then it restarts the task selection algorithm.
- Downsides:
  - Stations will make more requests, going from "one per round" to "one per task".
  - Since the Station does not know how much time is left in the current round, it can still perform a retrieval check that will finish in the next round and thus won't be rewarded.

WDYT?

As I wrote my comment, it seems to me that the alternative I described is better. It's also less reliant on the central party - in the future, Stations can listen to RoundStarted events directly by querying the RPC API.

bajtos: I removed the round `deadline` in [550b263](https://github.com/filecoin-station/spark-api/pull/80/commits/550b263991278282aaba029cf25eb1f6fffdae0c).

We can add it back in the future if (and only if) needed.

bajtos: And [46712c5](https://github.com/filecoin-station/spark-api/pull/80/commits/46712c5db93f52cea8dbc75646e060dae5ab7a5a)

bajtos: And [cfa9afa](https://github.com/filecoin-station/spark-api/pull/80/commits/cfa9afa4a68b9c9ee41b6631dc4e121ac452b676) 🙈

lib/round-tracker.js
juliangruber: Thank you!

juliangruber (approved):
Great work, Miro! 👏 👏

lib/round-tracker.js
juliangruber: I agree with your reasoning, removing deadlines is a good move!

PR #82, opened 2023-09-12
feat: decouple retrieval results from retrievals

bajtos:
1. Decouple `retrieval_results` table from `retrieval` to allow us to submit results for retrieval tasks with no linked `retrieval` row.

2. Rename `retrieval_results` to `measurements` to align with MERidian vocabulary.

3. Update `setRetrievalResult` to support the new data schema. This allows existing SPARK checkers to keep reporting measurements.

4. Remove `GET /retrieval/:id` endpoint and replace it with a new endpoint `GET /measurements/:id`. This endpoint is used by integration tests in SPARK checker, we will need to update these tests after this PR lands.

5. Create a new endpoint for submitting retrieval results based on per-round tasks - `POST /measurements`.

I am envisioning the following workflow for the SPARK checker node:
1. Get the current round & the tasks via `GET /rounds/current` - see #80
2. Pick a task, perform the retrieval
3. Call the new API `POST /measurement` to submit the result
4. Rinse and repeat

Connect to https://github.com/filecoin-station/spark/issues/13

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,8 +12,8 @@ const handler = async (req, res, client, getCurrentRound) => {
     await setRetrievalResult(req, res, client, Number(segs[1]), getCurrentRound)
   } else if (segs[0] === 'retrievals' && req.method === 'GET') {
     assert.fail(501, 'This API endpoint is no longer supported.')
-  } else if (segs[0] === 'retrieval-results' && req.method === 'GET') {
-    await getRetrievalResult(req, res, client, Number(segs[1]))
+  } else if (segs[0] === 'measurements' && req.method === 'GET') {
+    await getMeasurement(req, res, client, Number(segs[1]))
   } else if (segs[0] === 'rounds' && req.method === 'GET') {
     await getRoundDetails(req, res, client, getCurrentRound, segs[1])
   } else {
@@ -76,7 +76,7 @@ const setRetrievalResult = async (req, res, client, retrievalId, getCurrentRound
   validate(result, 'attestation', { type: 'string', required: false })
 
   const { rows } = await client.query(`
-      INSERT INTO retrieval_results (
+      INSERT INTO measurements (
         spark_version,
         zinnia_version,
         cid,
@@ -120,11 +120,11 @@ const setRetrievalResult = async (req, res, client, retrievalId, getCurrentRound
   if (!rows.length) {
     assert.fail(404, 'Retrieval Not Found')
   }
-  json(res, { retrievalResultId: rows[0].id })
+  json(res, { measurementId: rows[0].id })
 }
 
-const getRetrievalResult = async (req, res, client, retrievalResultId) => {
-  assert(!Number.isNaN(retrievalResultId), 400, 'Invalid RetrievalResult ID')
+const getMeasurement = async (req, res, client, measurementId) => {
+  assert(!Number.isNaN(measurementId), 400, 'Invalid RetrievalResult ID')
   const { rows: [resultRow] } = await client.query(`
     SELECT
       id,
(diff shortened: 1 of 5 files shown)


test/test.js
```diff
@@ -475,16 +451,16 @@ describe('Routes', () => {
...
       await assertResponseStatus(res, 200)
       const body = await res.json()
-      assert.strictEqual(body.id, retrievalId)
+      assert.strictEqual(body.id, retrievalResultId)
       assert.strictEqual(body.cid, cid)
       assert.strictEqual(body.providerAddress, providerAddress)
       assert.strictEqual(body.protocol, protocol)
       assert.strictEqual(body.sparkVersion, '1.2.3')
       assert.strictEqual(body.zinniaVersion, '2.3.4')
-      assert(body.createdAt)
```
bajtos, line 487: In the new data schema, we don't have "retrieval created at", as this information won't be available in the new SPARK checker version.

We can use `finishedAt` field instead - this requires updating all Grafana charts.

Thoughts?

test/test.js
```diff
@@ -356,7 +358,10 @@ describe('Routes', () => {
       )
       await assertResponseStatus(res, 200)
     })
-    it('ignores duplicate submissions', async () => {
+
+    // Duplicate submissions are ok, because we filter out duplicates in the fraud detection step
+    it('allows duplicate submissions', async () => {
```
bajtos, line 363: Do we have any concerns about DoS attacks? Before my change, clients could submit only as many results as many measurements they created. After my change, they can remove as many measurements as they like. I think my change is not making the situation significantly worse than it was before.

juliangruber:
> Decouple retrieval_results table from retrieval to allow us to submit results for retrieval tasks with no linked retrieval row

I wonder if we even still need the `retrievals` table, now that we have `retrieval_tasks`. Would it make sense to merge `retrievals` into `retrieval_tasks`?

juliangruber:
> Remove GET /retrieval/:id endpoint and replace it with a new endpoint GET /retrieval-results/:id. These endpoints are used by integration tests in SPARK checker, we will need to update these tests after this PR lands.

Ideally (I think) we'll keep `/retrievals/:id` and make that serve retrieval tasks 🤔

bajtos:
@juliangruber I updated the PR based on our IRL discussion:

- Renamed `retrieval_results` to `measurements`
- Renamed `GET /retrieval-results/:id` to `GET /measurements/:id`

I also added a new endpoint `POST /measurements` to be used by the updated SPARK module.

I think this PR is going to break our Grafana charts, I'll update the queries to use the new table name `measurements` after this PR is landed.

test/test.js
juliangruber: Sgtm!

test/test.js
juliangruber: What do you mean by "clients can remove as many measurements as they like"?

index.js
```diff
@@ -11,7 +11,11 @@ const handler = async (req, res, client, getCurrentRound) => {
   } else if (segs[0] === 'retrievals' && req.method === 'PATCH') {
     await setRetrievalResult(req, res, client, Number(segs[1]), getCurrentRound)
```
juliangruber, line 12: Can we remove this route completely? SPARK clients have to update anyway and aren't rewarded yet, so I think we can prioritize the code cleanup now already

juliangruber (changes requested):

test/test.js
bajtos: Sorry, that was a typo. I meant:
_After my change, they can submit as many measurements as they like._

index.js
juliangruber: Got it, ok!

juliangruber, line 12:
```suggestion
    // TODO: Deprecate once clients have been updated
    await setRetrievalResult(req, res, client, Number(segs[1]), getCurrentRound)
```

test/test.js
juliangruber: I think it's ok. We can discuss multiple strategies once we are more concerned with production deployment topics, or once we have many clients

juliangruber (approved):

PR #84, opened 2023-09-18
fix: allow up to 250 concurrent spark-api connections

bajtos:
Trying to fix the following problem found in logs:

> Instance 080e126b565608 reached hard limit of 25 concurrent
> connections. This usually indicates your app is not closing
> connections properly or is not closing them fast enough for
> the traffic levels it is handling. Scaling resources, number
> of instances or increasing your hard limit might help.

### DIFF
--- a/fly.toml
+++ b/fly.toml
@@ -19,8 +19,8 @@ processes = []
   protocol = "tcp"
   script_checks = []
   [services.concurrency]
-    hard_limit = 25
-    soft_limit = 20
+    hard_limit = 250
+    soft_limit = 200
     type = "connections"
 
   [[services.ports]]


fly.toml
```diff
@@ -19,8 +19,8 @@ processes = []
   protocol = "tcp"
   script_checks = []
   [services.concurrency]
-    hard_limit = 25
-    soft_limit = 20
+    hard_limit = 250
+    soft_limit = 200
```
bajtos, line 23: @juliangruber, do you remember why you chose these particular limits? I think we should be able to handle even more than 250 req/s, maybe I can bump this up to something like 2500/2000?

juliangruber: I didn't choose these limits, they were part of the default template created by the fly cli. Changing them is most likely the right step for us

juliangruber (approved):

bajtos:
Since we are running two SPARK-API instances (2x250 connections) and access the DB from Grafana (up to tens of concurrent connections), I bumped up the max connections limit in the DB to 1000.

```
spark-api/spark-db ❯ flyctl postgres config update --max-connections 1000
NAME           	VALUE	TARGET VALUE	RESTART REQUIRED
max-connections	300  	1000        	true

? Are you sure you want to apply these changes? Yes
Performing update...
Update complete!
Please note that some of your changes will require a cluster restart before they will be applied.
? Restart cluster now? Yes
Identifying cluster role(s)
  Machine 2874de7f654328: replica
  Machine e28657eea65e68: leader
Restarting machine 2874de7f654328
  Waiting for 2874de7f654328 to become healthy (started, 3/3)
Machine 2874de7f654328 restarted successfully!
Attempting to failover e28657eea65e68
Restarting machine e28657eea65e68
  Waiting for e28657eea65e68 to become healthy (started, 3/3)
Machine e28657eea65e68 restarted successfully!
Postgres cluster has been successfully restarted!
```

PR #85, opened 2023-09-18
fix: healthcheck config

bajtos:
- Increase grace period to 10s, we are fetching FEVM state at startup

- Configure the healthcheck URL as `/rounds/current`. After my recent changes, the root URL `/` is returning 404, and thus failing the check.


juliangruber:
👏

PR #88, opened 2023-09-18
perf: add index for `measurements.finished_at`

bajtos:
Many Grafana charts are filtering and grouping using this column, I hope the index will improve performance of those queries.


migrations/017.do.index-measurements-by-time.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX CONCURRENTLY measurements_finished_at ON measurements (finished_at);
```
juliangruber, line 1: 👍

PR #94, opened 2023-09-26
Add telemetry

juliangruber:
Closes https://github.com/filecoin-station/spark-evaluate/issues/2

See also https://github.com/filecoin-station/spark-evaluate/pull/15

### DIFF
--- a/spark-publish/index.js
+++ b/spark-publish/index.js
@@ -1,6 +1,7 @@
 /* global File */
 
 import timers from 'node:timers/promises'
+import { record } from './lib/telemetry.js'
 
 export const publish = async ({
   client,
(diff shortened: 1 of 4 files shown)


spark-publish/index.js
```diff
@@ -37,20 +38,24 @@ export const publish = async ({
...
+  const uploadMeasurementsDuration = new Date() - start
   logger.log(`Measurements packaged in ${cid}`)
 
   // Call contract with CID
   logger.log('ie.addMeasurements()...')
+  start = new Date()
   const tx = await ieContract.addMeasurements(cid.toString())
   const receipt = await tx.wait()
   const event = receipt.events.find(e => e.event === 'MeasurementsAdded')
   const { roundIndex } = event.args
```
bajtos, line 57: Nice! What do you think about updating `measurements` with the `roundIndex` value we got from the smart-contract interaction? That would solve the problem we discussed on Slack, where we end up with a part of measurements belonging to the previous round because of a race condition/timing issues.

bajtos (approved):
:shipit:

spark-publish/index.js
bajtos: I mean writing `roundIndex` to Postgres DB.

juliangruber: If we need that association, here seems to be a good place to do it 👍

PR #98, opened 2023-10-02
test: remove delay after all tests are done

bajtos:
Before my change, there was a delay of approx. two seconds after all tests complete but before the Mocha process exits. This delay was caused by the recently added InfluxDB client (commit 35176074).

This commits modifies the test suite to explicitly close the InfluxDB client after all tests finished.


juliangruber (approved):
awesome!

PR #101, opened 2023-10-11
feat: `GET /rounds/meridian/:address/:round`

bajtos:
Introduce a new API for obtaining SPARK round details for the given Meridian contract address + IE round.

Links:
- https://github.com/filecoin-station/spark/issues/13

PR #102, opened 2023-10-11
Revert "feat: `GET /round/meridian/:address/:round` (#101)"

bajtos:
This reverts commit 79e7407e056e279ce74702b29c9283e5f319a056 (#101)

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,8 +17,6 @@ const handler = async (req, res, client, getCurrentRound) => {
     await createMeasurement(req, res, client, getCurrentRound)
   } else if (segs[0] === 'measurements' && req.method === 'GET') {
     await getMeasurement(req, res, client, Number(segs[1]))
-  } else if (segs[0] === 'rounds' && segs[1] === 'meridian' && req.method === 'GET') {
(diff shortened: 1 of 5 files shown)


bajtos:
> Why did you revert this?

See https://github.com/filecoin-station/spark-api/pull/101#discussion_r1355132177

I should have cross-posted that information in the PR description 🙈 

The previous version required an expensive database migration that took longer than Fly timeouts allow. I re-ran the GHA deployment step thinking was maybe temporary, but it wasn't.

I am also becoming a bit concerned about the size of our database and the growth rate. The variant proposed in #101 would make this a bit worse. The new version implemented in #103 stores much less data.

bajtos:
Also, because the database migration failed, we ended up with a broken deployment:
- One Fly instance was probably waiting for the migration to finish, not serving requests yet
- The other Fly instance was up and running, but could not insert data into the DB because of the mismatch between the SQL schema assumed by the Node.js code and the actual DB schema in Postgres

As a result, I was seeing many errors in the logs.

So I reverted the change ASAP to restore working state.

juliangruber:
Thank you for the context! All sounds good

PR #103, opened 2023-10-11
feat: `GET /round/meridian/:address/:round`

bajtos:
Introduce a new API for obtaining SPARK round details for the given Meridian contract address + IE round.

This is a reworked version of #101 that is much lighter on DB storage & migration.

Links:
- https://github.com/filecoin-station/spark/issues/13

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,6 +17,8 @@ const handler = async (req, res, client, getCurrentRound) => {
     await createMeasurement(req, res, client, getCurrentRound)
   } else if (segs[0] === 'measurements' && req.method === 'GET') {
     await getMeasurement(req, res, client, Number(segs[1]))
+  } else if (segs[0] === 'rounds' && segs[1] === 'meridian' && req.method === 'GET') {
+    await getMeridianRoundDetails(req, res, client, segs[2], segs[3])
   } else if (segs[0] === 'rounds' && req.method === 'GET') {
     await getRoundDetails(req, res, client, getCurrentRound, segs[1])
   } else {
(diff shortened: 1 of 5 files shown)


lib/round-tracker.js
```diff
@@ -141,8 +141,8 @@ export async function mapCurrentMeridianRoundToSparkRound ({
     // this edge case by telling Postgres to ignore conflicts (`ON CONFLICT DO NOTHING)`
     await pgClient.query(`
       INSERT INTO meridian_contract_versions
-      (contract_address, spark_round_offset, last_spark_round_number)
-      VALUES ($1, $2, $3)
+      (contract_address, spark_round_offset, last_spark_round_number, first_spark_round_number)
+      VALUES ($1, $2, $3, $3)
```
bajtos, line 145: This is the first SPARK round managed by this IE contract version, therefore first and last spark round numbers are the same.

index.js
```diff
@@ -237,15 +239,19 @@ const getRoundDetails = async (req, res, client, getCurrentRound, roundParam) =>
...
   if (!round) {
     return notFound(res)
   }
 
-  const { rows: tasks } = await client.query('SELECT * FROM retrieval_tasks WHERE round_id = $1', [roundNumber])
+  const { rows: tasks } = await client.query('SELECT * FROM retrieval_tasks WHERE round_id = $1', [round.id])
 
   json(res, {
-    roundId: roundNumber.toString(),
+    roundId: round.id.toString(),
```
juliangruber, line 254: why did this change from `roundNumber` to `round.id`?

juliangruber (approved):

index.js
bajtos: I was re-using the code from https://github.com/filecoin-station/spark-api/pull/101 where this change was necessary.

It's not required for my PR, but I find it nicer that we are using the round number from the DB round row we fetched. If we ever need to make changes similar to #111 here again, then the code will be ready.

PR #104, opened 2023-10-12
feat: record "inet_group" in measurements

bajtos:
Convert the IP address of the SPARK checker node submitting a measurement into an IPv4 /24 subnet (e.g. 127.0.0.0).

Recognize `Fly-Client-IP` and `CF-Connecting-IP` request headers to obtain the real client address as observed by Fly's/CloudFlare's load balancer.

Requests coming via IPv6 (or other transports that IP) get assigned a hard-coded group `ipv6`.

Links:
- https://github.com/filecoin-station/spark/issues/29

### DIFF
--- a/lib/inet-grouping.js
+++ b/lib/inet-grouping.js
@@ -1,0 +1,24 @@
+// See https://stackoverflow.com/a/36760050/69868
+const IPV4_REGEX = /^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.){3}(25[0-5]|(2[0-4]|1\d|[1-9]|)\d)$/
+
+export const mapRequestToInetGroup = (/** @type {import('node:http').IncomingMessage} */ req) => {
+  let addr = req.socket.remoteAddress
+
+  const flyClientAddr = req.headers['fly-client-ip']
+  if (flyClientAddr) addr = flyClientAddr
+
+  if (addr?.startsWith('::ffff:')) {
+    // Some operating systems are wrapping the IPv4 address into IPv6
+    // https://www.apnic.net/get-ip/faqs/what-is-an-ip-address/ipv6-address-types/
+    addr = addr.slice(7)
+  }
+
+  if (!IPV4_REGEX.test(addr)) {
+    return undefined
+  }
+
+  // Hide the last byte of the IPv4 address, change it to "0"
+  addr = addr.slice(0, addr.lastIndexOf('.')) + '.0'
+
+  return addr
+}
--- a/lib/inet-grouping.js
+++ b/lib/inet-grouping.js
@@ -1,11 +1,108 @@
+import { base64url } from 'multiformats/bases/base64'
+
 // See https://stackoverflow.com/a/36760050/69868
 const IPV4_REGEX = /^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.){3}(25[0-5]|(2[0-4]|1\d|[1-9]|)\d)$/
 
-export const mapRequestToInetGroup = (/** @type {import('node:http').IncomingMessage} */ req) => {
+/**
+ * @param {import('pg').Client} client
+ * @param {import('node:http').IncomingMessage} req
+ * @param {Date} [now]
+ * @returns {string}
+ */
+export const mapRequestToInetGroup = async (pgClient, req, now = new Date()) => {
+  const subnet = mapRequestToSubnet(req)
+  const group = await mapSubnetToInetGroup(pgClient, subnet, now)
+  return group
+}
+
+/**
+ * @param {import('pg').Client} client
+ * @param {string} subnet
+ * @param {Date} [now]
+ * @returns {Promise<string>}
+ */
+export const mapSubnetToInetGroup = async (pgClient, subnet, now = new Date()) => {
+  const { rows: [found] } = await pgClient.query(
+    'SELECT id FROM inet_groups WHERE subnet = $1',
+    [subnet]
+  )
+  if (found) return found.id
+
+  let remainingRetries = 5
+  for (;; remainingRetries--) {
+    const group = generateUniqueGroupId(now)
+    try {
+      // There is a race condition: between the time we looked up the existing group
+      // and the time we execute this query, a handler for a different request from
+      // the same subnet may have already defined the group for this subnet.
+      // Solution: when insert fails with a conflict in the group id, use that existing group
+      // instead of the value we generated ourselves.
+      //
+      // ON CONFLICT DO UPDATE is needed to tell PG to return the existing id on conflict
+      // ON CONFLICT DO NOTHING would not return the existing id on conflict
+      const { rows: [created] } = await pgClient.query(`
+        INSERT INTO inet_groups (id, subnet, created_at)
+        VALUES ($1, $2, $3)
+        ON CONFLICT (subnet) DO UPDATE SET id = inet_groups.id
+        RETURNING id
+       `, [
+        group,
+        subnet,
+        now.toISOString()
+      ])
+      return created.id
+    } catch (err) {
+      if (err.code === 23505 && err.constraint === 'inet_groups_pkey') {
+        // Retry with a different random id
+        if (remainingRetries > 0) continue
+        // Fail if we exhausted allowed attempts
+        throw new Error('Failed to generate a unique group id')
+      }
+      throw err
+    }
+  }
+}
+
+/**
+ * @param {Date} [now]
+ * @returns {Promise<string>}
+ */
+const generateUniqueGroupId = (now) => {
+  const data = new ArrayBuffer(8)
+
+  const prefix = new Uint8Array(data, 0, 2)
+
+  // To reduce the likelihood of generating the same group values for different subnets,
+  // and also to make it easier to diagnose invalid use of groups (e.g. after they should
+  // have been expired), let's prefix the group with a single byte derived from the number
+  // of hours elapsed since the Unix epoch.
+  // 251 is the largest prime number that fits into a single byte.
+  prefix[0] = now.getTime() / 3600_000 % 251
+
+  // We are running this code in more than one process. To reduce possible clashes, let's
+  // add another byte derived from the millisecond time.
+  prefix[1] = now.getTime() % 251
+
+  // Finally, add six random bytes
+  globalThis.crypto.getRandomValues(new Uint8Array(data, 2, 6))
+
+  return base64url.encode(new Uint8Array(data))
+}
+
+/**
+ * @param {import('node:http').IncomingMessage} req
+ * @returns {Promise<string>}
+ */
+export const mapRequestToSubnet = (req) => {
   let addr = req.socket.remoteAddress
 
   const flyClientAddr = req.headers['fly-client-ip']
   if (flyClientAddr) addr = flyClientAddr
+
+  // TODO accept the value in this header only when the request is coming from Cloudflare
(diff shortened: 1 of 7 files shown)


bajtos:
Notes & discussion points:

I decided to use a more generic field name "inet_group" instead of something like "ipv4_subnet", because I think we will most likely need to iterate on this mechanism. For example, we may want to map (some) addresses to [ASNs])(https://www.cloudflare.com/en-gb/learning/network-layer/what-is-an-autonomous-system/).

**Privacy aspects of collecting IPv4 subnet info**

(1) CCPA

Quoting from https://iapp.org/news/a/are-ip-addresses-personal-information-under-ccpa:

> The CCPA’s proposed regulations
> (...)
> [I]f a business collects the IP addresses of visitors to its websites but does not link the IP address to any particular consumer or household, and could not reasonably link the IP address with a particular consumer or household, then the IP address would not be ‘personal information.

I am not a lawyer, but I think a /24 subnet (a group of 255 possible IP addresses) does not create a link to a particular consumer or household, simply because there can be 255 different people/households included in that group.

(2) GDPR

I could not find information whether we can use the same reasoning as above.

However, I believe we can claim we have a _legitimate interest_ to collect IPv4 subnets since we are relying on this information to prevent fraud.

Quoting from https://law.stackexchange.com/a/77814 (bold emphasis is mine):
> IP addresses are personal data. That means you need a _legal basis_ to process them, but not necessarily consent from the user.
> (...)
> As the answers in the questions you linked indicate, there are alternative legal bases to consent. The GDPR offers a choice of six legal bases (Art 6(1) lit a–f). In most cases, you would instead rely on a “legitimate interest” for logging. But it's not enough to claim that you have a legitimate interest. You must have a clear purpose for which such logs would be necessary, and you would then have to weigh this legitimate interest against the interests, rights, and freedoms of the affected data subjects. **If such logs are necessary for security and anti-abuse purposes, your legitimate interest test is likely to prevail.**

lib/inet-grouping.js
```diff
@@ -0,0 +1,24 @@
...
+export const mapRequestToInetGroup = (/** @type {import('node:http').IncomingMessage} */ req) => {
+  let addr = req.socket.remoteAddress
+
+  const flyClientAddr = req.headers['fly-client-ip']
+  if (flyClientAddr) addr = flyClientAddr
+
+  if (addr?.startsWith('::ffff:')) {
+    // Some operating systems are wrapping the IPv4 address into IPv6
+    // https://www.apnic.net/get-ip/faqs/what-is-an-ip-address/ipv6-address-types/
+    addr = addr.slice(7)
```
juliangruber, line 13: can we rely on there either being an ipv4 address or a wrapped ipv4 address, but never an ipv6 address?

lib/inet-grouping.js
```diff
@@ -0,0 +1,24 @@
+// See https://stackoverflow.com/a/36760050/69868
+const IPV4_REGEX = /^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.){3}(25[0-5]|(2[0-4]|1\d|[1-9]|)\d)$/
+
+export const mapRequestToInetGroup = (/** @type {import('node:http').IncomingMessage} */ req) => {
+  let addr = req.socket.remoteAddress
```
juliangruber, line 5: can be undefined, see https://nodejs.org/api/net.html#socketremoteaddress

lib/inet-grouping.js
```diff
@@ -0,0 +1,24 @@
...
+    // https://www.apnic.net/get-ip/faqs/what-is-an-ip-address/ipv6-address-types/
+    addr = addr.slice(7)
+  }
+
+  if (!IPV4_REGEX.test(addr)) {
+    return undefined
+  }
+
+  // Hide the last byte of the IPv4 address, change it to "0"
+  addr = addr.slice(0, addr.lastIndexOf('.')) + '.0'
```
juliangruber, line 21: This way we might be more flexible in the future, if we need to change the strategy

```suggestion
  addr = addr.slice(0, addr.lastIndexOf('.')) + '.0/24'
```

juliangruber (changes requested):

bajtos:
Related discussions:

- https://github.com/filecoin-station/proj-mgmt/issues/55
- https://github.com/filecoin-station/spark-api/pull/56

lib/inet-grouping.js
bajtos: No, we cannot. I am using IPV4_REGEX to detect & reject regular IPv6 addresses; see a few source-code lines below.

bajtos:
I improved the pull request as follows:

1. Instead of recording the actual IPv4 subnet value in the measurements, we map subnets to randomly generated `inet_group` strings.

2. We need to synchronise subnet->group mapping between SPARK processes (we are running two processes in parallel). We discusses using Redis, but I decided to use our DB instead. I feel adding Redis would increase the complexity of our architecture and operations, which I don't want to.

Caveats we may want to address in the future:

1. Currently, the subnet's inet_group value does not change over time. This could be used to track user's activity over longer period of time. To avoid that, we should expire group IDs - e.g. by periodically deleting all rows older than 24 hours.

2. We collect and store subnet values in our DB, which we (SPARK operators) can access. This allows us to reverse-map a measurement back to the IPv4 subnet it was submitted from. It is unclear whether this setup requires us to get consent from Station users. The good part is that this data is private; we are not sharing it with anybody else.

3. We rely on Cloudflare to provide IPv4 access only. This is problematic for two reasons:
   1. Cloudflare is pushing for IPv6 adoption. It is not possible to disable IPv6 via the web dashboard, I had to invoke their REST API. I am not sure how long this hack will last.
   2. There may be legitimate users that run Stations on IPv6-only networks. We should figure out how to support them. 
   
   FWIW, quoting from https://developer.apple.com/support/ipv6/
   > As of June 1, 2016, all apps submitted to the App Store must support IPv6-only networks.

Depending on the priorities and our bandwidth, instead of making incremental improvements for this design (and dig our hole a bit deeper), we may want to prioritise implementing a decentralised "proof of IPv4 address" instead, see https://github.com/filecoin-station/spark/issues/20

Having said that, I feel the version proposed in this pull request is good enough to get us started. We still need to implement the fraud-detection step leveraging `inet_group` in spark-evaluate.

lib/inet-grouping.js
```diff
@@ -1,24 +1,122 @@
...
+ */
+const generateUniqueGroupId = (now) => {
+  const data = new ArrayBuffer(8)
+
+  const prefix = new Uint8Array(data, 0, 2)
+
+  // To reduce the likelihood of generating the same group values for different subnets,
+  // and also to make it easier to diagnose invalid use of groups (e.g. after they should
+  // have been expired), let's prefix the group with a single byte derived from the number
+  // of hours elapsed since the Unix epoch.
```
juliangruber, line 79: How is this helping with conflicts? The number of hours elapsed doesn't change for 1h, so all calls in one 30 minutes round see the same hour, right? Or are you worried about conflicts over a long time?

Can we instead use a source of randomness with enough entropy (or long enough group ids), so that we don't need to factor in time or worry about conflicts? Eg something like `sha256(extractSubnet(ip))`?

lib/inet-grouping.js
```diff
@@ -1,24 +1,122 @@
...
+ * @param {import('node:http').IncomingMessage} req
+ * @returns {Promise<string>}
+ */
+export const mapRequestToSubnet = (req) => {
   let addr = req.socket.remoteAddress
 
   const flyClientAddr = req.headers['fly-client-ip']
   if (flyClientAddr) addr = flyClientAddr
 
+  // TODO accept the value in this header only when the request is coming from Cloudflare
```
juliangruber, line 103: How can we solve this?

juliangruber (changes requested):

lib/inet-grouping.js
bajtos: I guess I am too cautious here.

- We need groups to be unique only within each round. It's acceptable to assign the same group to one subnet in round X and a different subnet in round Y. (But I don't think it's a good idea, as it makes troubleshooting more difficult.)
- With 6 random bytes, we have 256^6 different group IDs. There are 256^3 different subnets we are mapping to these groups. With a truly random generator, it would take more than 16 million resets (256^3) before we start recycling the group IDs.
- We are also using `UNIQUE` constraint in our DB to detect duplicate IDs.

The only edge case I can think of is this:
- We happen to reset our mapping table in the middle of a round
- We generate a sequence of random bytes that was already used before the reset (very unlikely, but still has a non-zero probability)
- As a result, two different subnets are mapped to the same group id in this round

The prefix "hours % 251" is handling this edge case if we assume that we will periodically (once a day?) delete group ids older than 24 hours.

> The number of hours elapsed doesn't change for 1h, so all calls in one 30 minutes round see the same hour, right?

Yes, the number does not change for the same round (unless we reset the mapping).

That's why I added the second byte based on milliseconds since epoch.

> Can we instead use a source of randomness with enough entropy

I am using [crypto.getRandomValues()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) - I believe that's the most robust random number generator we have available in JavaScript.

---

OK, I think I over-engineered this part. I am going to remove the time but keep using 8 bytes of random data (encoded as 12 characters). That should be good enough to get us started. We can easily tweak this later based on the data we observe.

lib/inet-grouping.js
bajtos: See https://developers.cloudflare.com/support/troubleshooting/restoring-visitor-ips/restoring-original-visitor-ips/, which explains how to configure `mod_remoteip` module for Apache to correctly restore original visitor IP:

> Define trusted proxy addresses by creating /etc/apache2/conf-available/remoteip.conf by entering the following code and [ Cloudflare IPs](https://www.cloudflare.com/ips/):
>
> ```
> RemoteIPHeader CF-Connecting-IP
> RemoteIPTrustedProxy 192.0.2.1 (example IP address)
> RemoteIPTrustedProxy 192.0.2.2 (example IP address)
> (repeat for all Cloudflare IPs listed at https://www.cloudflare.com/ips/)
> ```

I improved the code comment to explain, see https://github.com/filecoin-station/spark-api/pull/104/commits/480d770f89cc3dc42bf159524dc985c925e5ce7f

lib/inet-grouping.js
bajtos: See https://github.com/filecoin-station/spark-api/pull/104/commits/d9f0a40b1501ce41742c0d4feb7c88017f27be6d

juliangruber: Sounds good, let's watch for `UNIQUE` constraint violations to see when our randomness isn't good enough?

bajtos: > Sounds good, let's watch for `UNIQUE` constraint violations to see when our randomness isn't good enough?

We are kind of already doing that - when the UNIQUE constraint is violated, we generate a new random id, retrying up to 5 times. If all five attempts fail, then we abort with the following error:

https://github.com/filecoin-station/spark-api/pull/104/files#diff-2ce02f2adf56d846dab8a6c0282e7e21ac77403f11f58c89e08e2d18ff8c15bdR60

```js
throw new Error('Failed to generate a unique group id')
```

This error will be reported to Sentry, so we should be good here.

juliangruber: perfect!

bajtos: > This error will be reported to Sentry, so we should be good here.

Ouch, I take that back. We are not logging request-handling errors to Sentry, only unhandled errors crashing the process.

Fixing it here: https://github.com/filecoin-station/spark-api/pull/112

PR #109, opened 2023-10-18
add redirect to the right domain

juliangruber:
We need to disallow access to `spark.fly.dev`, since it doesn't sit behind cloudflare

### DIFF
--- a/bin/spark.js
+++ b/bin/spark.js
@@ -12,7 +12,7 @@ import assert from 'node:assert'
 const {
   PORT = 8080,
   HOST = '127.0.0.1',
-  DOMAIN = 'api.filspark.com',
+  DOMAIN = 'localhost',
   DATABASE_URL,
   SENTRY_ENVIRONMENT = 'development'
 } = process.env
--- a/bin/spark.js
+++ b/bin/spark.js
@@ -12,6 +12,7 @@ import assert from 'node:assert'
 const {
   PORT = 8080,
   HOST = '127.0.0.1',
+  DOMAIN = 'api.filspark.com',
   DATABASE_URL,
   SENTRY_ENVIRONMENT = 'development'
 } = process.env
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -4,7 +4,10 @@ import getRawBody from 'raw-body'
 import assert from 'http-assert'
 import { validate } from './lib/validate.js'
 
-const handler = async (req, res, client, getCurrentRound) => {
+const handler = async (req, res, client, getCurrentRound, domain) => {
+  if (req.headers.host.split(':')[0] !== domain) {
```
juliangruber, line 8: disregard the port, so that tests still work (where we don't know the port beforehand)

bajtos (approved):

bin/spark.js
```diff
@@ -12,6 +12,7 @@ import assert from 'node:assert'
 const {
   PORT = 8080,
   HOST = '127.0.0.1',
+  DOMAIN = 'api.filspark.com',
```
bajtos, line 15: On the second thought, maybe we should configure this via Fly config file and don't assume any default domain for people running this via `npm start`?

bajtos: Alternatively, define domain as `localhost` by default, so that we can assume domain is always configured.

Then add an environment variable to our Fly config to change this to the domain used by our deployment.

juliangruber: Good point, I hadn't thought about dev mode

juliangruber: Let's default to `localhost` and set an env var

test/test.js
```diff
@@ -669,6 +670,38 @@ describe('Routes', () => {
...
+        const spark = `http://127.0.0.1:${server.address().port}`
+        const res = await fetch(
+          `${spark}/rounds/${currentSparkRoundNumber}`,
+          { redirect: 'manual' }
+        )
+        await assertResponseStatus(res, 301)
+        assert.strictEqual(res.headers.get('location'), `https://foobar/rounds/${currentSparkRoundNumber}`)
+      } finally {
+        server.closeAllConnections()
+        server.close()
```
bajtos, line 700: It makes me wonder: instead of closing the server at the end, would it be enough to call `server.unref()` right after we create it?

bajtos (approved):
:shipit:

PR #114, opened 2023-10-24
fix: rework spark-api Dockerfile

bajtos:
Remove dependency on Volta, use Node.js official Docker image instead.

Unify the content of spark-api and spark-publish Dockerfiles
so that we consistently use the same setup everywhere.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,37 +1,43 @@
-FROM debian:bullseye as builder
+# syntax = docker/dockerfile:1
 
+# Adjust NODE_VERSION as desired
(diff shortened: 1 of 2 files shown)


bajtos:
This fixes the problem with Volta as observed here: https://github.com/filecoin-station/spark-api/actions/runs/6627795309/job/18003528229

It should also speed up deploys because we can reuse the cached Node.js base layer instead of downloading it via Volta.

juliangruber (commented):
+1

I had just copy/pasted that from Fly

PR #125, opened 2023-11-06
feat: bump TASKS_PER_ROUND to 200

bajtos:
Now that we have over 1900 retrieval templates (after we land #124), we can increase the number of tasks per round to the desired level.

SPARK nodes wait for 10 seconds between checks. With a 30-minute round length, a node can perform at most `30*60/10 = 180` tasks per round.

In practice, we will always get some nodes that don't perform all tasks, e.g. because retrieval takes longer or the node is not online for the entire duration of the round.

To make it easier to observe how our system handles such situations, I am intentionally choosing `TASKS_PER_ROUND` value higher than the maximum theoretical limit, to ensure we encounter this edge case very often.

Links:
- https://github.com/filecoin-station/spark/issues/12
- #124


juliangruber (approved):

bajtos:
😍 

Screenshot 2023-11-06 at 11 20 56

PR #126, opened 2023-11-06
feat: remove POST & PATCH `/measurements`

bajtos:
These endpoints have been superseded by `POST /measurements` as part of https://github.com/filecoin-station/spark/issues/13

After this change, spark-api returns 410 OUTDATED CLIENT errors when these endpoints are called.

This removal allowed me to further clean up our code:
- remove `retrievals` table
- delete `retrieval_templates` that were soft-deleted
- drop the column `retrieval_templates.deleted`

THIS PR MUST BE LANDED AFTER #43 OTHERWISE DATABASE MIGRATIONS WILL BREAK.

Resolves https://github.com/filecoin-station/spark-api/issues/73

PR #127, opened 2023-11-06
fix: remove `retrieval_templates.deleted`

bajtos:
Fix the filename of the migration script that prevented it from being executed.

See https://github.com/filecoin-station/spark-api/pull/126


juliangruber (commented):
👍

PR #131, opened 2023-11-13
test: use the constant TASKS_PER_ROUND

bajtos:
So that we don't have to change the tests when we tweak the number of tasks chosen for each round.

This is a follow-up for 86c3299d6, bb878bfc8 and 0c307518d.


juliangruber (approved):
Thank you!

PR #137, opened 2023-11-16
feat: report load of spark-publish

bajtos:
Modify spark-publish to include another telemetry field: load factor.  I.e. what fraction of measurements waiting to be published we were able to process.

The load factor should be under 100% most time. If it only occasionally goes over 100% then the service should be able to eventually self-heal.  The trouble starts when we are consistently over 100% - the backlog of unpublished measurements is growing faster than we can process it.


juliangruber (approved):
👏

PR #138, opened 2023-11-20
fix: formula for spark-publish load

bajtos:
We want to measure totalCount / maxMeasurements.

This is a follow-up for #137.


bajtos:
Data coming from the old formula:

Screenshot 2023-11-20 at 07 59 17

bajtos:
@juliangruber hindsight please

PR #139, opened 2023-11-20
fix: count of unpublished measurements

bajtos:
Count all rows only after we collected the rows to publish.  spark-api adds more measurements in between our queries. If we query the total count first and rows to publish second, we would report total count smaller than the count of published rows.

See #138


bajtos:
Example of a misleading log:

```
Publishing 8030 measurements. Total unpublished: 3897. Batch size: 60000.
```

bajtos:

juliangruber (commented):
👍

PR #143, opened 2023-11-21
add caching for GET /rounds/current

juliangruber:
After merging, adjust https://dash.cloudflare.com/37573110e38849a343d93b727953188f/filspark.com/caching/cache-rules/7cc5c26ebc3c400c9666c5a0f83ac50d to read the `cache-control` header again

### DIFF
--- a/index.js
+++ b/index.js
@@ -141,9 +141,10 @@ const getMeasurement = async (req, res, client, measurementId) => {
 const getRoundDetails = async (req, res, client, getCurrentRound, roundParam) => {
   const roundNumber = await parseRoundNumberOrCurrent(getCurrentRound, roundParam)
 
-  if (roundParam === 'current') {
-    res.setHeader('cache-control', 'no-store')
-  }
+  res.setHeader(
+    'cache-control',
+    `max-age=${roundParam === 'current' ? 60 : 31536000}}`
+  )
 
   await replyWithDetailsForRoundNumber(res, client, roundNumber)
 }


index.js
```diff
@@ -141,9 +141,10 @@ const getMeasurement = async (req, res, client, measurementId) => {
 const getRoundDetails = async (req, res, client, getCurrentRound, roundParam) => {
   const roundNumber = await parseRoundNumberOrCurrent(getCurrentRound, roundParam)
 
-  if (roundParam === 'current') {
-    res.setHeader('cache-control', 'no-store')
-  }
+  res.setHeader(
+    'cache-control',
+    `max-age=${roundParam === 'current' ? 60 : 31536000}}`
```
juliangruber, line 146: I thought about a dynamic setting for the current round, which ensures it's not longer than the expected round end.  Wdyt @bajtos?

bajtos: This is a good idea.

In https://github.com/filecoin-station/spark-api/pull/147, I reworked "get tasks for the current round" logic so that we don't need dynamic max-age any more.

However, we should eventually move the logic of determining "what's the current meridian round index" from spark-api to spark-checker. I can imagine that it can be useful for spark-checker nodes to not start any new work (retrieval) if the current round is coming to an end. 

By the time we finish the retrieval, submit the measurement to spark-api, and spark-publisher commits the measurement to Meridian, the new round will have already started, and our measurement will be evaluated as invalid (measuring a task that does not belong to this round).

I think we should make the check robust to handle changes in round lengths. Ideally, the smart contract should indicate when the current round is expected to end. I think we already have this value in Meridian state - the public field `currentRoundEndBlockNumber`.

I am proposing to open a new GH issue to track this idea. WDYT?

juliangruber:
Closing in favor of https://github.com/filecoin-station/spark-api/pull/147

bajtos:
> After merging, adjust https://dash.cloudflare.com/37573110e38849a343d93b727953188f/filspark.com/caching/cache-rules/7cc5c26ebc3c400c9666c5a0f83ac50d to read the `cache-control` header again

Done.

Issue #144, opened 2023-11-21
Our database is melting down under heavy load


bajtos:
Grafana dashboards are now taking ages to load. Queries executed by spark-api are gradually becoming slower and slower. 

1. Setup monitoring to understand how DB performance changes over time
2. Identify opportunities for "quick win" performance improvements and apply them

Future improvements:
- https://github.com/filecoin-station/spark-api/issues/140

bajtos:
Todo:
- [ ] finish https://github.com/filecoin-station/spark-evaluate/pull/67
- [ ] rework all Grafana visualisations to source data from InfluxDB 
- [ ] periodically delete old rounds & measurements from our DB - keep data for last 7 days

bajtos:
> periodically delete old rounds & measurements from our DB - keep data for last 30 days

This isn't urgent, I opened a follow-up issue to implement it later:
- https://github.com/filecoin-station/spark-api/issues/154

bajtos:
More optimisation opportunities:

- Create an index to support the query we use to visualise SPARK rounds

  ```sql
  SELECT created_at as time, id as round_number
  FROM spark_rounds
  WHERE "created_at" BETWEEN '2023-11-27T20:12:13.471Z' AND '2023-11-28T08:12:13.471Z'
  ```

PR #147, opened 2023-11-22
perf: better caching of current round details

bajtos:
Rework the implementation of `GET /rounds/current` to redirect to a permalink `GET /rounds/meridian/{addr}/{round}`. This allows us to cache better the data queried from our database.

1. The redirection logic does not need the database; it builds the new location using in-memory data.

   In the future, we can rework spark-checker to determine the current round locally by querying the on-chain state of the Meridian contract.

2. The response for the permalink for a given meridian round is immutable and, therefore, can be cached forever. If all works out as expected, we should see only one request per round, and the CDN cache will take care of the rest.

I have also improved the database schema to store the meridian contract address and round index in `spark_rounds` in order to enable fast lookup of Spark rounds from Meridian rounds,

This PR supersedes and closes #143

Links:
- https://github.com/filecoin-station/spark-api/issues/144

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,7 +33,8 @@ const handler = async (req, res, client, getCurrentRound, domain) => {
 }
 
 const createMeasurement = async (req, res, client, getCurrentRound) => {
-  const round = await getCurrentRound()
+  console.log('round', getCurrentRound())
+  const { sparkRoundNumber } = getCurrentRound()
   const body = await getRawBody(req, { limit: '100kb' })
   const measurement = JSON.parse(body)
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -33,7 +33,8 @@ const handler = async (req, res, client, getCurrentRound, domain) => {
 }
 
 const createMeasurement = async (req, res, client, getCurrentRound) => {
-  const round = await getCurrentRound()
+  console.log('round', getCurrentRound())
```
juliangruber, line 36:
```suggestion
```

juliangruber (approved):
great work!

PR #151, opened 2023-11-24
Integrate new web3.storage. Closes #141

### LINKED ISSUE #141: Migrate to the new web3.storage API
See https://mailchi.mp/web3.storage/major-changes-to-web3storage?e=77416530cd

> what this means for you, a current web3.storage user. Here’s the tl;dr:
> - The new upload API and client is live. Its account system is completely separate from that of the legacy web3.storage.
> - The legacy web3.storage upload API, pinning service API, and w3name will sunset on January 9, 2024. Until then, you will still be able to upload to your legacy account up to your free account limit. The old dashboard can be found [here](https://old.web3.storage/account/).
> - All uploads to the legacy web3.storage prior to January 9, 2024 will continue to be stored until further notice, though over time latency and availability may degrade.

/cc @juliangruber FYI

juliangruber:
- [x] Matching `spark-evaluate` PR https://github.com/filecoin-station/spark-evaluate/pull/70

Files are now published to the root entry of the CID (vs at `/measurements.json`), therefore this PR and the `spark-evaluate` PR need to be merged together.

Closes #141

### DIFF
--- a/spark-publish/index.js
+++ b/spark-publish/index.js
@@ -53,7 +53,7 @@ export const publish = async ({
     'measurements.json',
     { type: 'application/json' }
   )
-  const cid = await web3Storage.put([file])
+  const cid = await web3Storage.uploadFile(file)
   const uploadMeasurementsDuration = new Date() - start
   logger.log(`Measurements packaged in ${cid}`)
 
--- a/spark-publish/bin/spark-publish.js
+++ b/spark-publish/bin/spark-publish.js
@@ -3,7 +3,10 @@ import { startPublishLoop } from '../index.js'
 import { IE_CONTRACT_ABI, IE_CONTRACT_ADDRESS, RPC_URL } from '../ie-contract-config.js'
 import Sentry from '@sentry/node'
 import assert from 'node:assert'
-import { Web3Storage } from 'web3.storage'
+import * as Client from '@web3-storage/w3up-client'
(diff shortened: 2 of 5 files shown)


spark-publish/index.js
```diff
@@ -53,7 +53,7 @@ export const publish = async ({
     'measurements.json',
     { type: 'application/json' }
   )
-  const cid = await web3Storage.put([file])
+  const cid = await web3Storage.uploadFile(file)
```
juliangruber, line 56: I noticed a change in behavior here, see https://filecoinproject.slack.com/archives/C027XP5BTGB/p1700818334024649

juliangruber: This new one is actually simpler, since we don't need to support the extra complexity of unixfs

bajtos (approved):

juliangruber:
> Can you please grep through the source code and remove any remaining places that may be using `WEB3_STORAGE_API_TOKEN`?

There are no string matches for this, were you thinking about something particular?

bajtos:
> > Can you please grep through the source code and remove any remaining places that may be using `WEB3_STORAGE_API_TOKEN`?
> 
> There are no string matches for this, were you thinking about something particular?

Not really. Just to make sure we don't accidentally leave behind some dead code. Thanks for checking!

Issue #150, opened 2023-11-22
We have run out of log storage allowance


bajtos:
https://papertrailapp.com/dashboard
> Your account has reached 96% log data transfer for this period.

The projected volume is 86.1GB/month. A plan with 7 days of searchable logs would cost us $425/monthly. That feels quite a lot to me.

Action items:

- [ ] Setup an alert so that we are notified about this problem in the future
- [x] Review our logs and look for options to reduce the load. For example, does spark-api need to log every incoming HTTP request?

Links:
- https://github.com/filecoin-station/spark-api/issues/108

bajtos:
Maybe we can start by configuring our Fly log forwarder to ignore logs from spark-api - they are noise and very few of them are meaningful IMO.

See https://fly.io/docs/going-to-production/monitoring/exporting-logs/#shipping-specific-logs

patrickwoodhead:
Can we close this issue @juliangruber ?

juliangruber:
>  Setup an alert so that we are notified about this problem in the future

I believe this part is still missing.

<img width="603" alt="Screenshot 2024-01-03 at 13 32 41" src="https://github.com/filecoin-station/spark-api/assets/10247/b807d9b8-3a09-4c2f-b94e-92d510ed2112">

I think this is the closest we can get from PT. @bajtos are you ok enabling this option?

bajtos:
I believe Additional Usage should be already enabled. If it's not then please do enable it.

juliangruber:
You're right. Then this should be all we can do about this issues 🙏

PR #153, opened 2023-11-27
Disable request logging in prod

juliangruber:
#150

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ WORKDIR /app
 ENV NODE_ENV production
 ENV SENTRY_ENVIRONMENT production
 ENV DOMAIN api.filspark.com
+ENV REQUEST_LOGGING false
 
 #######################################################################
 # Throw-away build stage to reduce size of final image
--- a/bin/spark.js
+++ b/bin/spark.js
@@ -15,7 +15,8 @@ const {
   HOST = '127.0.0.1',
   DOMAIN = 'localhost',
   DATABASE_URL,
-  SENTRY_ENVIRONMENT = 'development'
+  SENTRY_ENVIRONMENT = 'development',
+  REQUEST_LOGGING = 'true'
 } = process.env
 
 const pkg = JSON.parse(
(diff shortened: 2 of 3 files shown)


bajtos:
> Log requests but ignore in PT: I doubt request logs will be useful to anyone at this volume, aggregates stats in Influx are the way to go

I found it useful to see live request logs to get a feel for how much traffic is coming, how long is request handling duration, and what are the responses. Before this change, I could do this by running `fly logs -a spark` or visiting https://fly.io/apps/spark/monitoring

Let's keep your current solution and wait if I ever need ☝🏻 badly enough to warrant a different solution.

I was originally thinking to follow [Shipping Specific Logs](https://fly.io/docs/going-to-production/monitoring/exporting-logs/#shipping-specific-logs) to ignore spark-api logs, but I see the config does not allow exclusions nor a list of services. We would have to look for a more involved solution, and that's probably not worth the effort right now.

bajtos:
> Pass a separate request logger as an option: `createHandler({ requestLogger: noop })`. That removes the conditions from our `createHandler` function, but otoh it feels weird to have a separate logger only for one log type

Yeah, it would be nice to get rid of those conditions.

Have you considered adding a new method `log.request` to the existing logger?

```js
const logger = Object.create(console)
// alternatively:
// const logger = {...console}
logger.request = requestLoggerEnabled
  ? console.log
  : ()=>{}
```

juliangruber:
> I found it useful to see live request logs to get a feel for how much traffic is coming, how long is request handling duration, and what are the responses. Before this change, I could do this by running fly logs -a spark or visiting https://fly.io/apps/spark/monitoring

If you did this in prod right before I landed this PR, your terminal would fill up horizontally in <1s. I don't think that gives you any meaningful information. Unless you're a very fast interpreter 😎 

> I was originally thinking to follow [Shipping Specific Logs](https://fly.io/docs/going-to-production/monitoring/exporting-logs/#shipping-specific-logs) to ignore spark-api logs, but I see the config does not allow exclusions nor a list of services. We would have to look for a more involved solution, and that's probably not worth the effort right now.

The advantage of this PR is that we still get error logs from `spark-api`, just not request logs.

> Have you considered adding a new method log.request to the existing logger?

Good idea! Will follow up

juliangruber:
https://github.com/filecoin-station/spark-api/pull/159

PR #160, opened 2023-11-30
fix: preserve date values set to `null`

bajtos:
Before this change, when the client sent `end_at: null`, we would convert that value to `Date(null)`, which is the Unix epoch.

After this change, we preserve null values as null.

Links:
- https://github.com/filecoin-station/spark/issues/43

### DIFF
--- a/index.js
+++ b/index.js
@@ -93,10 +93,10 @@ const createMeasurement = async (req, res, client, getCurrentRound) => {
     measurement.protocol,
     measurement.participantAddress,
     measurement.timeout || false,
-    new Date(measurement.startAt),
+    parseDateStringOrNull(measurement.startAt),
     measurement.statusCode,
-    new Date(measurement.firstByteAt),
-    new Date(measurement.endAt),
+    parseDateStringOrNull(measurement.firstByteAt),
+    parseDateStringOrNull(measurement.endAt),
     measurement.byteLength,
     measurement.attestation,
     inetGroup,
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -213,6 +213,43 @@ describe('Routes', () => {
...
+
+      const { rows: [measurementRow] } = await client.query(`
+          SELECT *
+          FROM measurements
+          WHERE id = $1
+        `, [
+        id
+      ])
+
+      assert.strictEqual(measurementRow.status_code, null)
```
bajtos, line 249: `status_code` is NULLABLE in the database. When the measurement does not specify the status code at all (value is `undefined`), the roundtrip through the database converts the value to `null`.

This PR does not change the existing behaviour, just adds a test to document it.

🤷🏻‍♂️

index.js
```diff
@@ -285,3 +285,8 @@ export const createHandler = async ({
       })
   }
 }
+
+const parseDateStringOrNull = (strOrNull) => {
+  if (strOrNull === undefined || strOrNull === null) return undefined
```
juliangruber, line 290: slightly confusing: It's called `OrNull` but returns `undefined` and accepts `null` or `undefined`

juliangruber: What about `parseDateStringOrUndefined`?

bajtos: I decided to use `parseOptionalDate`

juliangruber (approved):

PR #165, opened 2023-12-07
switch to NDJSON

juliangruber:
Deploy after https://github.com/filecoin-station/spark-evaluate/pull/87

### DIFF
--- a/spark-publish/index.js
+++ b/spark-publish/index.js
@@ -50,7 +50,7 @@ export const publish = async ({
   let start = new Date()
   const file = new File(
     [measurements.map(m => JSON.stringify(m)).join('\n')],
-    'measurements.json',
+    'measurements.ndjson',
     { type: 'application/json' }
   )
   const cid = await web3Storage.uploadFile(file)
--- a/spark-publish/index.js
+++ b/spark-publish/index.js
@@ -49,7 +49,7 @@ export const publish = async ({
   // Share measurements
   let start = new Date()
   const file = new File(
-    [JSON.stringify(measurements)],
+    [measurements.map(m => JSON.stringify(m)).join('\n')],
     'measurements.json',
     { type: 'application/json' }
   )
--- a/spark-publish/test/test.js
+++ b/spark-publish/test/test.js
@@ -87,7 +87,7 @@ describe('integration', () => {
   it('publishes', async () => {
     await client.query('DELETE FROM measurements')
 
-    const measurementRecorded = {
+    const measurements = [{
       sparkVersion: '1.2.3',
       zinniaVersion: '0.5.6',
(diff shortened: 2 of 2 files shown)


spark-publish/index.js
```diff
@@ -49,7 +49,7 @@ export const publish = async ({
   // Share measurements
   let start = new Date()
   const file = new File(
-    [JSON.stringify(measurements)],
+    [measurements.map(m => JSON.stringify(m)).join('\n')],
     'measurements.json',
```
bajtos, line 53:
```suggestion
    'measurements.ndjson',
```

This could also make it easier to detect old (JSON) vs new (NDJSON) format, if we ever need it.

bajtos: Ah, I see you choose a different approach to support both JSON & NDJSON.

I still think we should use a different file extension, because the file we produce is not a valid JSON.

spark-publish/test/test.js
```diff
@@ -201,13 +220,16 @@ describe('integration', () => {
...
+    assert.strictEqual(payload.length, 2)
     const published = payload[0]
-    assert.strictEqual(published.spark_version, measurementRecorded.sparkVersion)
-    assert.strictEqual(published.cid, measurementRecorded.cid)
-    assert.strictEqual(published.inet_group, measurementRecorded.inetGroup)
-    assert.strictEqual(published.car_too_large, measurementRecorded.carTooLarge)
+    assert.strictEqual(published.spark_version, measurements[0].sparkVersion)
+    assert.strictEqual(published.cid, measurements[0].cid)
+    assert.strictEqual(published.inet_group, measurements[0].inetGroup)
+    assert.strictEqual(published.car_too_large, measurements[0].carTooLarge)
```
bajtos, line 232: How about adding

```js
const measurementRecorded = measurements[0]
```

to avoid code churn in these assertions?

spark-publish/index.js
```diff
@@ -49,7 +49,7 @@ export const publish = async ({
   // Share measurements
   let start = new Date()
   const file = new File(
-    [JSON.stringify(measurements)],
+    [measurements.map(m => JSON.stringify(m)).join('\n')],
```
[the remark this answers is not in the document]
juliangruber: I wasn't trying to help with memory in the publisher, only the consumer. We can optimize this yes, but with https://github.com/filecoin-station/spark-api/pull/166 I don't see a need for it yet. And the buffer approach has an issue, as you noticed, with needing to know the buffer size beforehand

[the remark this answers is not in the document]
bajtos: Fair enough 👍🏻

bajtos (approved):
:shipit:

PR #166, opened 2023-12-07
Move `publish()` to own CLI, for memory

juliangruber:
We have two CLIs now: Once that runs the loop, and one that performs one iteration of the loop.

This way we shouldn't have any long-running memory gc issues

### DIFF
--- a/spark-publish/bin/publish-batch.js
+++ b/spark-publish/bin/publish-batch.js
@@ -1,0 +1,64 @@
+// Run `publish()` in a short lived script, to help with memory issues
+
+import Sentry from '@sentry/node'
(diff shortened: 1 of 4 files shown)


bajtos (approved):
NOICE!

Do you know if the InfluxDB client will manage to flush the write queue before the process exits? If not, then we will loose telemetry.

I guess the best process is to land this PR as is, observe what happens in production, and then tweak the InfluxDB integration as needed.

juliangruber:
Really good catch! Afaik it won't wait. I've added this behavior manually 🙏

bajtos:
👏🏻

PR #172, opened 2023-12-12
feat: maxTasksPerNode in round details

bajtos:
Add a new per-round configuration parameter allowing us to tweak how many retrieval requests are performed by each checker node.

The concept of "tasks per node" is described in [SPARK tasking v2](https://www.notion.so/pl-strflt/SPARK-tasking-v2-604e26d57f6b4892946525bcb3a77104?pvs=4#ded1cd98c2664a2289453d38e2715643).

Links:
- https://github.com/filecoin-station/roadmap/issues/52

### DIFF
--- a/index.js
+++ b/index.js
@@ -172,6 +172,7 @@ const replyWithDetailsForRoundNumber = async (res, client, roundNumber) => {
 
   json(res, {
     roundId: round.id.toString(),
+    maxTasksPerNode: round.max_tasks_per_node,
     retrievalTasks: tasks.map(t => ({
       cid: t.cid,
       providerAddress: t.provider_address,
@@ -204,6 +205,7 @@ const getMeridianRoundDetails = async (_req, res, client, meridianAddress, merid
   res.setHeader('cache-control', `public, max-age=${ONE_YEAR_IN_SECONDS}, immutable`)
   json(res, {
     roundId: round.id.toString(),
+    maxTasksPerNode: round.max_tasks_per_node,
     retrievalTasks: tasks.map(t => ({
       cid: t.cid,
       providerAddress: t.provider_address,
--- a/lib/round-tracker.js
+++ b/lib/round-tracker.js
@@ -10,6 +10,9 @@ import { createMeridianContract } from './ie-contract.js'
 // We will need to tweak this value based on measurements; that's why I put it here as a constant.
 export const TASKS_PER_ROUND = 4000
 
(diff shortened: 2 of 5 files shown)


lib/round-tracker.js
```diff
@@ -192,13 +195,14 @@ export async function maybeCreateSparkRound (pgClient, {
 }) {
   const { rowCount } = await pgClient.query(`
     INSERT INTO spark_rounds
-    (id, created_at, meridian_address, meridian_round)
-    VALUES ($1, now(), $2, $3)
+    (id, created_at, meridian_address, meridian_round, max_tasks_per_node)
```
juliangruber, line 198: is this PR missing a migration file? 🤔

bajtos: Good catch! Added in c04a1fc

index.js
```diff
@@ -172,6 +172,7 @@ const replyWithDetailsForRoundNumber = async (res, client, roundNumber) => {
 
   json(res, {
     roundId: round.id.toString(),
+    maxTasksPerNode: round.max_tasks_per_node,
```
juliangruber, line 175: would it be more useful to return `delayBetweenRetrievals` (in ms) here? Since we know the round length, we can let the client be simpler and calculate this here

juliangruber (changes requested):

index.js
bajtos: It was my initial thought, too. But: while `delayBetweenRetrievals` is simpler in the short-term, it's something we will need to rework in the longer term as we progress towards decentralised protocol as described in our design docs.

Also spark-api knows as little about the round length as spark-checker does. 

In the future, I'd like spark-checkers to query on-chain state to determine when a new round starts and how many epochs/blocks remain in the current round. That will produce a much more accurate value than spark-api can ever provide.

I am imagining that in each iteration, spark-checker can:
1. Find out what's the current round index and when the next round is should start (see [`currentRoundEndBlockNumber`](https://github.com/Meridian-IE/impact-evaluator/blob/156ad2506b99db9e951f4444384e1a94958ea1cd/src/ImpactEvaluator.sol#L9))
2. Determine how many tasks we have completed this round.
3. If too many - then wait for the next round before doing any more work
4. If less than maxTasksPerNode, then do another retrieval
5. Decide how long to wait before the next iteration depending on how much time is remaining in the current round and how many more tasks we can run until we reach the max

juliangruber (approved):

PR #173, opened 2023-12-13
feat: reject measurements from spark versions <1.7.0

bajtos:
Links:
- https://github.com/filecoin-station/roadmap/issues/52

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@ import getRawBody from 'raw-body'
 import assert from 'http-assert'
 import { validate } from './lib/validate.js'
 import { mapRequestToInetGroup } from './lib/inet-grouping.js'
+import { satisfies } from 'compare-versions'
 
 const handler = async (req, res, client, getCurrentRound, domain) => {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -4,6 +4,7 @@ import getRawBody from 'raw-body'
 import assert from 'http-assert'
 import { validate } from './lib/validate.js'
 import { mapRequestToInetGroup } from './lib/inet-grouping.js'
+import { satisfies } from 'compare-versions'
```
bajtos, line 7: While [semver](https://www.npmjs.com/package/semver) is a popular package and it's maintained by the npm team, it has 93kB and depends on [lru-cache](https://www.npmjs.com/package/lru-cache). 

[compare-versions](https://www.npmjs.com/package/compare-versions) has zero dependencies and only 55kB.

juliangruber: I'm happy to try the competitor module, but if it has bugs I would immediately first to replace it with `semver`, as bundle size doesn't matter much for our services imo

juliangruber (approved):

PR #174, opened 2023-12-13
Implement new RPC_URL rotation strategy

juliangruber:
Stick with one JSON-RPC provider until it fails. This has the advantage that on downtime we don't keep retrying the same provider every other request


bajtos (approved):
:shipit:

PR #183, opened 2024-01-03
Set `published_as` and `commitments` in one transaction

juliangruber:
For data consistency


spark-publish/index.js
```diff
@@ -66,21 +66,30 @@ export const publish = async ({
...
-  await client.query(`
-    UPDATE measurements
-    SET published_as = $1
-    WHERE id = ANY($2::int[])
-  `, [
-    cid.toString(),
-    measurements.map(m => m.id)
-  ])
+  try {
+    await client.query('BEGIN')
```
bajtos, line 70: This does not work. See https://node-postgres.com/features/transactions

> You must use the same client instance for all statements within a transaction. PostgreSQL isolates a transaction to individual clients. This means if you initialize or use transactions with the pool.query method you will have problems. Do not use transactions with the pool.query method.

This is how I am have solved this problem elsewhere:
https://github.com/filecoin-station/spark-api/blob/ca10df39450cf50ff68be871869b763287bc0939/lib/round-tracker.js#L33-L47

I would like to eventually introduce a helper function for running queries in a transaction, but maybe it's too early if we need it in two places only.

The idea:

```js
// usage
await runInTransaction(pgPool, (client) => {
 await client.query()
 await client.query()
})

// implementation
const runInTransaction = (pgPool, fn) => {
  const pgClient = await pgPool.connect()
  try {
    await pgClient.query('BEGIN')
    await fn(pgClient)
    await pgClient.query('COMMIT')
   } catch (err) {
    await pgClient.query('ROLLBACK')
    throw err
   } finally {
     pgClient.release()
  }
}

bajtos: Unless we can be sure that `client` is always `pg.Client` and never `pg.Pool`?

juliangruber: We're passing a pg pool here indeed. 

> I would like to eventually introduce a helper function for running queries in a transaction, but maybe it's too early if we need it in two places only.

I was wondering about this too, it seems strange that `pg` doesn't offer this. I think it would be nice to introduce this

bajtos (approved):
LGTM 👍🏻

juliangruber:
Since https://github.com/filecoin-station/spark-api/pull/187 we're not going to store `published_as` any more, so this PR is obsolete

bajtos:
> Since #187 we're not going to store `published_as` any more, so this PR is obsolete

🤔 

Here is the error/edge case to consider:
- The first SQL query to delete published records is executed
- The second SQL query to add a new row to `commitments` fails

We end up with measurements published for evaluation but not accessible from data in our DB.

Maybe this is unlikely enough that we don't need to worry about it.

juliangruber:
Indeed, we need this functionality for the `commitments` table. Will revive this PR

bajtos (approved):
:shipit:

PR #187, opened 2024-01-08
feat: drop `measurements.published_as` column

bajtos:
The column `measurements.published_as` is no longer needed because we are deleting measurements immediately after they are published. The index `measurements_not_published` is not needed either.

We are no longer querying `measurements` using a range of `finished_at` values, plus there will be at most ~50k measurements at any time; therefore, I feel we no longer need the index `measurements_finished_at`.

Before changing the schema, the migration script calls `TRUNCATE TABLE measurements` to ensure the migration is fast. At the time of opening this pull request, the `measurements` table was using 36 GB of disk space. This truncation is also a cleanup step to delete historical published measurements. In principle, it applies #186 to old measurements. Finally, by running `TRUNCATE`, we reclaim space used by the `measurements` table when it was keeping historical published measurements. (But now it is not.)

This is a follow-up for https://github.com/filecoin-station/spark-api/pull/186. I want to make these changes only after spark-publish was updated to delete published records.


spark-publish/index.js
```diff
@@ -66,13 +65,11 @@ export const publish = async ({
   const ieAddMeasurementsDuration = new Date() - start
   logger.log('Measurements added to round', roundIndex.toString())
 
-  // Mark measurements as shared
+  // Delete published measurements
```
bajtos, line 68: I had to include the commit from #186 in this PR to make the tests pass. I'll rebase this branch after #186 is landed to remove that commit.

The changes are best to be reviewed in this view:
https://github.com/filecoin-station/spark-api/pull/187/commits/ce3815b9e80057984bbc4e5f5163c2bfb0fbe3c6

juliangruber (approved):
great thinking to clean this up 🤩

PR #193, opened 2024-01-16
fix: set Fly primary region to CDG

bajtos:
Collocate the API service with our database.

After this PR is landed, we need to manually recreate Fly VM machines to move the service from IAD (Virginia/USA) to CDG (Paris, France).


juliangruber (approved):
Intersting, I thought we had used cdg for every deployment, but looks like here we might not have been asked?

PR #197, opened 2024-01-22
fix composite commitment key

juliangruber:
It is possible for two commitments to have the same CID (eg when the round was empty).

### DIFF
--- a/migrations/037.do.composite-commitment-key.sql
+++ b/migrations/037.do.composite-commitment-key.sql
@@ -1,0 +1,3 @@
+ALTER TABLE commitments
+  DROP CONSTRAINT commitments_pkey,
+  ADD PRIMARY KEY (cid, published_at);


bajtos (commented):
Can you please check how we are querying this table in spark-evaluate helper scripts? We need to ensure these queries hit the index and don't perform a full scan.

We may need to add `CREATE INDEX ON commitments (cid)` but I am not sure if it's needed.

juliangruber:
> Can you please check how we are querying this table in spark-evaluate helper scripts? We need to ensure these queries hit the index and don't perform a full scan.
> 
> We may need to add `CREATE INDEX ON commitments (cid)` but I am not sure if it's needed.

```
SELECT cid FROM commitments
  WHERE published_at >= $1 AND published_at < $2
```

and

```
SELECT cid FROM commitments
  WHERE published_at >= $1 AND published_at < $2
  ORDER BY published_at DESC
```

Wdyt?

bajtos:
Thanks for finding the queries. I think these queries don't need an index on CID, but let's not make assumptions and verify!

First, connect to our DB:

```
fly pg connect -a spark-db -d spark
```

Second, run EXPLAIN ANALYZE query:

```sql
EXPLAIN ANALYZE SELECT cid FROM commitments
  WHERE published_at >= now() - INTERVAL '7 days' 
  AND published_at < now() - INTERVAL '1 day'
```

Output tells us that PG used Index Scan - all is good 👍🏻 

```
 Index Scan using commitments_published_at on commitments  (cost=0.30..280.44 rows=324 width=60) (actual time=1.266..475
.626 rows=4121 loops=1)
   Index Cond: ((published_at >= (now() - '7 days'::interval)) AND (published_at < (now() - '1 day'::interval)))
 Planning Time: 0.766 ms
 Execution Time: 476.041 ms
```

The query plan for the second query is similar:

```
 Index Scan Backward using commitments_published_at on commitments  (cost=0.30..280.44 rows=324 width=68) (actual time=0.545..1.148 rows=4121 loops=1)
   Index Cond: ((published_at >= (now() - '7 days'::interval)) AND (published_at < (now() - '1 day'::interval)))
 Planning Time: 0.163 ms
 Execution Time: 1.382 ms
```

bajtos (approved):
:shipit:

juliangruber:
Thank you for verifying! 🙏

PR #207, opened 2024-02-07
feat: retrievable deals

bajtos:
Create a new table `retrievable_deals` and insert 10k `(cid, expires_at)` pairs to bootstrap the dataset.

Links:
- https://www.notion.so/pl-strflt/Moving-IPNI-queries-to-Spark-checkers-c58175c0f1c841e6bb82ac44a1472a98
- https://github.com/filecoin-station/spark/issues/40


migrations/039.do.retrievable-deals.sql
```diff
@@ -0,0 +1,10013 @@
+CREATE TABLE retrievable_deals (
+  cid TEXT NOT NULL PRIMARY KEY,
+  expires_at TIMESTAMPTZ NOT NULL
+);
```
bajtos, line 4: @juliangruber do you have any feedback on the schema definition and the table name in particular?

Quoting from the design doc:

> Introduce a new SQL table with a list of FIL+ LDN deals we want to test for retrievability. This table will be similar to `retrieval_templates` and have the following columns:
>     - cid (primary key; extracted from deal Label)
>     - expires_at (timestamp with tz, calculated from deal proposal’s EndEpoch field)
>     
> In the future, we will need metadata to match IPNI responses to this particular deal. We don’t have that implemented yet, so let's keep the db schema simple and the table size small for now.
>
> I considered adding `id` as a primary key, but we don’t use it anywhere. We can rework the primary key to include multiple columns after we add more columns. The benefit of the proposed scheme is that we will naturally enforce the uniqueness of templates. (The rows in `retrieval_templates` are not unique.)

juliangruber: `retrievable_deals` makes sense to me! The only alternative I can think of is something like `ipni_responses`

migrations/039.do.retrievable-deals.sql
```diff
@@ -0,0 +1,10013 @@
...
+
+-- First 10k deals to bootstrap the dataset. Ingest the remaining 22m deals manually.
+
+TRUNCATE TABLE retrievable_deals;
```
juliangruber, line 8: why truncate it first?

juliangruber (approved):

migrations/039.do.retrievable-deals.sql
bajtos: Ah, I copy&pasted the SQL generated by fil-deal-ingester. We don't need to truncate the table in this migration script, I'll remove that statement.

PR #209, opened 2024-02-08
feat: generate tasks from retrievable deals

bajtos:
Quoting from the design doc:

> Rework the code defining retrieval tasks for a new Meridian round to pick CIDs from the list of FIL+ LDN deals. Set columns `provider_address` and `protocol` to `NULL`.
>
>    *In the future, we will need metadata to match IPNI responses to this particular deal. We don’t have that yet, so let's keep the schema simple and the table small in size for now.*
>
> Verify that the response of `GET /rounds/meridian/{contract}/{round-index}` converts `NULL` values in `provider_address` and `protocol` to undefined values in JavaScript, therefore excluding these fields from the JSON response.
>    
> *The goal here is to preserve the provider address & protocol in the responses for older rounds where the retrieval tasks were defined with the provider address and protocol to use.*

Links:
- https://www.notion.so/pl-strflt/Moving-IPNI-queries-to-Spark-checkers-c58175c0f1c841e6bb82ac44a1472a98
- https://github.com/filecoin-station/spark/issues/40


bajtos:
This is a breaking change for Spark checker nodes. Nodes running the Spark version older than 1.8.0 will stop earning rewards.

I guess we should wait with this change until we modify spark-api to start rejecting outdated clients on Feb 15th.

juliangruber:
> This is a breaking change for Spark checker nodes. Nodes running the Spark version older than 1.8.0 will stop earning rewards.
> 
> I guess we should wait with this change until we modify spark-api to start rejecting outdated clients on Feb 15th.

Yeah now that we have communicated this I think we have to

lib/round-tracker.js
```diff
@@ -214,9 +214,10 @@ export async function maybeCreateSparkRound (pgClient, {
 
 async function defineTasksForRound (pgClient, sparkRoundNumber) {
   await pgClient.query(`
-    INSERT INTO retrieval_tasks (round_id, cid, provider_address, protocol)
-    SELECT $1 as round_id, cid, provider_address, protocol
-    FROM retrieval_templates
+    INSERT INTO retrieval_tasks (round_id, cid)
+    SELECT $1 as round_id, cid
+    FROM retrievable_deals
+    WHERE expires_at > now()
```
juliangruber, line 220: why was this condition added and not necessary before?

juliangruber (approved):

lib/round-tracker.js
bajtos: Great question!

There is no `expires_at` column in `retrieval_templates`.

bajtos:
We must not land this PR until https://github.com/filecoin-station/spark-api/pull/217 is landed first.

PR #213, opened 2024-02-10
fix turn `retrieval_results_id_seq` into bigint

juliangruber:
Fixes `error: nextval: reached maximum value of sequence "retrieval_results_id_seq" (2147483647)`


bajtos (commented):
👍🏻

PR #215, opened 2024-02-12
fix: 4x less tasks per round

bajtos:
spark-publish is not able to keep up with the number of measurements we are receiving. Let's reduce the number of tasks to allow the system to self-heal and give us more time to figure out how to scale this up.


bajtos:

juliangruber (commented):
👏

PR #257, opened 2024-04-25
Fix: assert stationId at 88 bytes instead of 64

PatrickNercessian:


PatrickNercessian:
Link: filecoin-station/roadmap#97

PR #259, opened 2024-04-30
fix accept f4 addresses

juliangruber:
When we added the address validation we hadn't realized that Station Core still submits f4 addresses as well. Fix in https://github.com/filecoin-station/core/pull/441, until the network all upgraded this patch is necessary


bajtos (commented):
👌🏻

PR #261, opened 2024-05-02
deps: upgrade Node.js in Dockerfile

bajtos:
It would be great to find a way how to automate this.
- Can Dependabot upgrade base image version for us?
- Alternatively, can we specify only the major version and let Docker pick the latest minor/patch version available at the time of building the image?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax = docker/dockerfile:1
 
 # Adjust NODE_VERSION as desired
-ARG NODE_VERSION=20.8.1
+ARG NODE_VERSION=20.12.2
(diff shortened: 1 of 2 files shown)


juliangruber (approved):

juliangruber:
> It would be great to find a way how to automate this.
> 
> * Can Dependabot upgrade base image version for us?
> * Alternatively, can we specify only the major version and let Docker pick the latest minor/patch version available at the time of building the image?

I don't know the answers to these questions. Dependabot upgrading this sounds great, however not pinning the version is risky - our deployments can break because of a node upgrade, and we won't know that that caused that

PR #280, opened 2024-05-10
feat: timeout & retry while waiting for FEVM tx

bajtos:
When waiting for confirmation of the `ie.addMeasurements()` transaction, implement a timeout of 2 minutes and retry scheme timing out after 10 minutes in total.

The goal is to prevent spark-publish from hanging indefinitely when the RPC API or the blockchain are not able to provide any confirmations.

Based on `add_measurements_duration_ms` collected in InfluxDB, most transactions are confirmed within 100 seconds, which is within the 2 minute timeout. There are occassional outliers going up to 500 seconds, they will be handled by the retries. We cannot afford to wait for the confirmation too long, otherwise we will not commit measurements before the round ends.

### DIFF
--- a/spark-publish/index.js
+++ b/spark-publish/index.js
@@ -66,10 +66,15 @@ export const publish = async ({
   start = new Date()
   const tx = await ieContract.addMeasurements(cid.toString())
   logger.log('Waiting for the transaction receipt:', tx.hash)
-  const receipt = await pRetry(() => tx.wait(1 /* confirmation(s) */, 120_000 /* 2 minutes */), {
-    onFailedAttempt: err => console.error(err),
-    maxRetryTime: 600_000 // 10-minute timeout
-  })
+  const receipt = await pRetry(
+    () => tx.wait(
+      1, // confirmation(s)
+      120_000 // 2 minutes
+    ), {
+      onFailedAttempt: err => console.error(err),
+      maxRetryTime: 600_000 // 10-minute timeout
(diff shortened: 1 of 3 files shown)


spark-publish/index.js
```diff
@@ -65,7 +66,10 @@ export const publish = async ({
   start = new Date()
   const tx = await ieContract.addMeasurements(cid.toString())
   logger.log('Waiting for the transaction receipt:', tx.hash)
-  const receipt = await tx.wait()
+  const receipt = await pRetry(() => tx.wait(1 /* confirmation(s) */, 120_000 /* 2 minutes */), {
+    onFailedAttempt: err => console.error(err),
+    maxRetryTime: 600_000 // 10-minute timeout
+  })
```
juliangruber, line 72: I find this a bit hard to read with the two inline comments, what do you think about this?

```suggestion
  const receipt = await pRetry(
    () => tx.wait(
      1, // confirmation(s)
      120_000 // 2 minutes
    ), {
      onFailedAttempt: err => console.error(err),
      maxRetryTime: 600_000 // 10-minute timeout
    }
  )
```

juliangruber (approved):
+1 to the approach, let's deploy and monitor the situation

spark-publish/index.js
bajtos: Yeah, I had similar concerns. Thank you for the suggestion!

PR #285, opened 2024-05-13
perf: stop counting rows after 10 millions

bajtos:
Apply https://github.com/filecoin-station/voyager-api/pull/28 from voyager-api

### DIFF
--- a/spark-publish/index.js
+++ b/spark-publish/index.js
@@ -44,8 +44,10 @@ export const publish = async ({
   // Fetch the count of all unpublished measurements - we need this for monitoring
   // Note: this number will be higher than `measurements.length` because spark-api adds more
   // measurements in between the previous and the next query.
+  // Note: counting ALL rows can put too much load on the DB server when the table is very large.
+  // Let's stop after we count 10 million rows. That's enough to let us know that we are in trouble.
   const totalCount = (await pgPool.query(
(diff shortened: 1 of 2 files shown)


juliangruber (approved):

bajtos:
I think this change is no longer needed after we added the index on `measurements.published_at` in #286.

```
EXPLAIN ANALYZE SELECT COUNT(*) FROM measurements;
                                                                                     QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize Aggregate  (cost=11340.10..11340.11 rows=1 width=8) (actual time=58.569..64.822 rows=1 loops=1)
   ->  Gather  (cost=11339.88..11340.09 rows=2 width=8) (actual time=58.187..64.814 rows=3 loops=1)
         Workers Planned: 2
         Workers Launched: 2
         ->  Partial Aggregate  (cost=10339.88..10339.89 rows=1 width=8) (actual time=54.277..54.278 rows=1 loops=3)
               ->  Parallel Index Only Scan using measurements_finished_at on measurements  (cost=0.42..9794.65 rows=218095 width=0) (actual time=0.038..43.680 rows=173951 loops=3)
                     Heap Fetches: 69034
 Planning Time: 0.397 ms
 Execution Time: 64.926 ms
(9 rows)
```

PR #290, opened 2024-05-13
feat: allow no roundStartEpoch for existing rounds

bajtos:
When the Glif RPC API calls are timing out, or when our system is broken and rounds are not advancing at all, allow the spark-api service to start, even though it cannot obtain the roundStartEpoch of the current round.


juliangruber (approved):
Let's discuss whether to land this PR, but it itself lgtm

bajtos:
I was thinking more about this problem and would like to change the architecture of our round tracking. These incremental changes seem a bit hacky and increase the complexity too much.

The first step on the new journey: https://github.com/filecoin-station/spark-api/pull/292

PR #292, opened 2024-05-14
feat: decouple round tracker from API routes

bajtos:
**feat: decouple round tracker from API routes**

Rework API handlers to get the current round from the database.

Benefits:

- This commit fixes the long-standing bug where the REST API returned different current round depending on which worker process received the request. This was caused by different timing when each worker process detected a new RoundStarted event.

- The new architecture allows us to further decouple the round tracker from the API handler. We can run it in background to not block process startup or even extract it into a standalone service.

Downsides:

- More database queries. Before my changes, the current round was available in memory. Now we need to read it from the database.  Affected endpoints:
  - `POST /measurements`
  - `GET /rounds/current`

Note: The next commit reworks the query recording a new measurement to obtain the current round number in a sub-query of the `INSERT` statement.

**perf: inline `getCurrentRound()`**

- Rework the query inserting new measurements to fetch the current round as part of the single INSERT statement.

- Inline `getCurrentRound()` implementation into the single remaining place calling it - the handler for `GET /rounds/current`.

### DIFF
--- a/bin/spark.js
+++ b/bin/spark.js
@@ -58,9 +58,10 @@ client.on('error', err => {
 })
 await migrate(client)
 
+console.log('Initializing round tracker...')
+const start = Date.now()
+
 try {
-  console.log('Initializing round tracker...')
-  const start = Date.now()
   const currentRound = await startRoundTracker(client)
   console.log(
     'Initialized round tracker in %sms. SPARK round number at service startup: %s',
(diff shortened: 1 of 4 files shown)


juliangruber (approved):
I very much agree with the approach of this PR 👏 While I'm also concerned about the extra queries on `POST /measurements`, I think we can optimize this later if we identify that there's a problem here. Great clean up!

PR #300, opened 2024-05-15
fix: Ethers v6 contract address getter

bajtos:
Fix the regression introduced by #264

### DIFF
--- a/lib/round-tracker.js
+++ b/lib/round-tracker.js
@@ -26,9 +26,9 @@ export const MAX_TASKS_PER_NODE = 15
 export async function startRoundTracker (pgPool) {
   const contract = await createMeridianContract()
 
-  const updateSparkRound = async (newRoundIndex) => {
+  const updateSparkRound = async (/** @type {bigint} */ newRoundIndex) => {
     const meridianRoundIndex = BigInt(newRoundIndex)
-    const meridianContractAddress = contract.address
+    const meridianContractAddress = await contract.getAddress()
 
     const roundStartEpoch = await getRoundStartEpoch(contract, meridianRoundIndex)
 
@@ -57,15 +57,22 @@ export async function startRoundTracker (pgPool) {
     }
   }
 
-  contract.on('RoundStart', (newRoundIndex) => {
+  const onRoundStart = (newRoundIndex) => {
     updateSparkRound(newRoundIndex).catch(err => {
       console.error('Cannot handle RoundStart:', err)
       Sentry.captureException(err)
     })
-  })
+  }
+  contract.on('RoundStart', onRoundStart)
+  const close = () => contract.removeListener('RoundStart', onRoundStart)
 
-  const currentRound = await updateSparkRound(await contract.currentRoundIndex())
-  return currentRound
+  try {
+    const currentRound = await updateSparkRound(await contract.currentRoundIndex())
+    return { ...currentRound, close }
+  } catch (err) {
+    close()
+    throw err
+  }
 }
 
 /*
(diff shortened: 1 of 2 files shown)


lib/round-tracker.js
```diff
@@ -26,9 +26,9 @@ export const MAX_TASKS_PER_NODE = 15
 export async function startRoundTracker (pgPool) {
   const contract = await createMeridianContract()
 
-  const updateSparkRound = async (newRoundIndex) => {
+  const updateSparkRound = async (/** @type {bigint} */ newRoundIndex) => {
     const meridianRoundIndex = BigInt(newRoundIndex)
-    const meridianContractAddress = contract.address
+    const meridianContractAddress = await contract.getAddress()
```
bajtos, line 31: This is the fix

lib/round-tracker.js
```diff
@@ -57,15 +57,22 @@ export async function startRoundTracker (pgPool) {
...
       console.error('Cannot handle RoundStart:', err)
       Sentry.captureException(err)
     })
-  })
-
-  const currentRound = await updateSparkRound(await contract.currentRoundIndex())
-  return currentRound
+  }
+  contract.on('RoundStart', onRoundStart)
+  const close = () => contract.removeListener('RoundStart', onRoundStart)
```
bajtos, line 67: I need this in order to allow the test to stop the tracker (unsubscribe from the events) after the test is done.

lib/round-tracker.js
```diff
@@ -57,15 +57,22 @@ export async function startRoundTracker (pgPool) {
...
+
+  try {
+    const currentRound = await updateSparkRound(await contract.currentRoundIndex())
+    return { ...currentRound, close }
+  } catch (err) {
+    close()
```
bajtos, line 73: When `await updateSparkRound()` throws, the caller does not have access to `close` and cannot unsubscribe.

To avoid leaking a background tracker, we must unsubscribe ourselves.

Since this is used by tests only, I think there is an alternative solution - remove all event listeners after each test. @juliangruber WDYT?

test/round-tracker.test.js
```diff
@@ -1,23 +1,27 @@
...
+  /** @type {pg.Pool} */
+  let pgPool
+  /** @type {pg.PoolClient} */
   let pgClient
 
   before(async () => {
-    pgClient = new pg.Client({ connectionString: DATABASE_URL })
-    await pgClient.connect()
+    pgPool = new pg.Pool({ connectionString: DATABASE_URL })
+    pgClient = await pgPool.connect()
```
bajtos, line 18: Existing tests use a single connection, I am not sure if they require that.

The new test needs `pgPool`, or at least that's what the type says.

To keep the changes minimal, I am changing the test setup to offer both `pgPool` and `pgClient`.

test/round-tracker.test.js
```diff
@@ -211,6 +215,15 @@ describe('Round Tracker', () => {
       assert.strictEqual(typeof startEpoch, 'number')
     })
   })
+
+  describe('startRoundTracker', () => {
+    it('detects the current round', async function () {
+      this.timeout(60_000)
```
bajtos, line 221: The pull request passed the CI but the deployment run timed out. I have increased the timeout in https://github.com/filecoin-station/spark-api/commit/4ba3f532d8b8e05f991d3e3bb718e12b056002f1

lib/round-tracker.js
juliangruber: What about we create `class RoundTracker`, on which you can call `start`, but also `close`? I believe we're outgrowing a single function here.

test/round-tracker.test.js
```diff
@@ -211,6 +215,15 @@ describe('Round Tracker', () => {
       assert.strictEqual(typeof startEpoch, 'number')
     })
   })
+
+  describe('startRoundTracker', () => {
```
juliangruber, line 219: 👏

PR #305, opened 2024-05-16
refactor: use AbortSignal instead of `close()`

bajtos:
Refactor cancellation of the background task created by `startRoundTracker` to use injected AbortSignal instead of returning a `close()` function.


juliangruber (approved):
👏

PR #315, opened 2024-05-22
publish: Add retry `ie.addMeasurements()`. Closes #293

### LINKED ISSUE #293: spark-publish: retry `.addMeasurements`
The publishing iteration often fails for intermittent RPC-API-related reasons that could be worked around by retrying the step committing the measurements batch CID.

- We incorrectly estimated the gas needed by the transaction and set the gas limit too low.
- Glif RPC API failed with 504 Gateway Time-out.
- (possibly more)

This does not affect the performance & correctness of the entire system (the next iteration will retry everything). However, it wastes resources.

- The DB query listing 100k measurements can take a long time to complete if we are under load and there are too many measurements waiting to be published.
- The data we upload to web3.storage becomes orphaned.

Let's implement a retry logic for the entire step "commit to the smart contract", i.e. a top-level retry for both `.addMeasurements` + `tx.wait()`.

Slack discussion: https://space-meridian.slack.com/archives/C06RPCL6QGL/p1715590945617159

juliangruber:
Closes #293

### DIFF
--- a/spark-publish/index.js
+++ b/spark-publish/index.js
@@ -65,7 +65,15 @@ export const publish = async ({
   // Call contract with CID
   const signal = AbortSignal.timeout(600_000) // 10-minute timeout
   const { roundIndex, ieAddMeasurementsDuration } = await pTimeout(
-    commitMeasurements({ cid, ieContract, logger, signal }),
+    pRetry(
+      () => commitMeasurements({ cid, ieContract, logger, signal }),
+      {
+        onFailedAttempt: err => console.error(err),
+        shouldRetry: err => err.code !== 'CALL_EXCEPTION',
+        signal,
+        retries: 5 // 5 * 2 minutes = 10 minutes - another measure to enforce ~10-minute timeout
+      }
+    ),
     { signal, milliseconds: Number.POSITIVE_INFINITY /* timeout is triggered by signal */ }
   )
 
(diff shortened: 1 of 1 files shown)


bajtos (changes requested):
We must keep `pRetry` here, because transactions can take more than 2 minutes to be confirmed by the network.

Desired flow:
1. create a transaction (`ieContract.addMeasurements`) only once
2. wait for the transaction to be confirmed, possibly multiple times

If we cannot create the transaction at all, then retry the entire operation.

However, if the transaction was created, then we must retry only `tx.wait`.

juliangruber:
You're absolutely right, what about this?

spark-publish/index.js
```diff
@@ -65,7 +65,14 @@ export const publish = async ({
   // Call contract with CID
   const signal = AbortSignal.timeout(600_000) // 10-minute timeout
   const { roundIndex, ieAddMeasurementsDuration } = await pTimeout(
-    commitMeasurements({ cid, ieContract, logger, signal }),
+    pRetry(
+      () => commitMeasurements({ cid, ieContract, logger, signal }),
+      {
+        onFailedAttempt: err => console.error(err),
```
bajtos, line 71: I think this will log some errors twice: the first time from `onFailedAttempt` configured for `pRetry(tx.wait)` and the second time from this place. However, errors thrown outside of pRetry will not be logged that way and we still need to log them here.

It would be great if we could detect the case when an error was already observed via the inner `onFailedAttempt` handler. I am not sure if we can use one of the additional error properties added by pRetry, e.g. `err.attemptNumber` or `err.retriesLeft` or whether we would have to add our own property in the `onFailedAttempt` handler.

Anyhow, this is a minor detail we can ignore for now and wait if it ever becomes a problem in production.

juliangruber: This only happens when the last attempt of the inner `p-retry` is exhausted, right?

juliangruber: I'm going to merge for now, let's continue discussion here 🙏

PR #319, opened 2024-05-27
fix: add `miner_id` to `retrievable_deals` PK

bajtos:
Before this change, we used `cid` as the primary key. As a result, we would have only one record for each CID.


bajtos:
I discovered this problem while troubleshooting https://filecoinproject.slack.com/archives/C075188L1R9

juliangruber (changes requested):
Very important fix 🙏 Could we please secure this through a regression test?

bajtos:
> Very important fix 🙏 Could we please secure this through a regression test?

Great idea! 💪🏻

bajtos:
Test added in [4dc65d1](https://github.com/filecoin-station/spark-api/pull/319/commits/4dc65d1090e7d87a70981ef2fda78f9eb1aad743)

@juliangruber LGTY now?

juliangruber (approved):
👏

PR #323, opened 2024-05-28
refactor: introduce monorepo-like layout

bajtos:
Introduce the following repository layout:

- `bin` contains scripts shared by all components, currently `bin/migrate.js`
- `migrations` contains SQL migrations scripts and `index.js` for running them
- each component has its own directory with its own fly.toml
  - logshipper
  - api
  - spark-db
  - publish
- there is a single Dockerfile shared by all workspaces
- `publish` and `api` are npm workspaces with their own `package.json` file
- the top-level `package.json` file is a npm workspace root


spark-api/fly.toml
```diff
@@ -10,6 +10,11 @@ processes = []
   PORT = "8080"
   HOST = "0.0.0.0"
   SENTRY_ENVIRONMENT = "production"
+  DOMAIN = "api.filspark.com"
+  REQUEST_LOGGING = "false"
+
+[build.args]
+  SERVICE=api
```
bajtos, line 17: https://fly.io/docs/reference/configuration/#specify-docker-build-arguments

spark-publish/fly.toml
```diff
@@ -1,4 +1,12 @@
 app = "spark-publish"
 primary_region = "cdg"
+
+[env]
+  SENTRY_ENVIRONMENT="production"
+  MAX_MEASUREMENTS_PER_ROUND="100000"
+
+[build.args]
+  SERVICE="publish"
```
bajtos, line 9: https://fly.io/docs/reference/configuration/#specify-docker-build-arguments

Dockerfile
```diff
@@ -40,5 +38,12 @@ FROM base
 # Copy built application
 COPY --from=build /app /app
 
-# Start the server by default, this can be overwritten at runtime
-CMD [ "npm", "run", "start" ]
+# Set to `publish` or `api`
```
juliangruber, line 41: What do you think about consistently only using `spark-api` or `api`? Currently it's mixed: On disk it's called `spark-api`, here the arg is `api`. What do you think about using `spark-api` everywhere, or just `api`? `spark-api` is the more explicit one, and matches the fly project name. `api` it's nicer, because there's less repetition (`spark-api/spark-api`)

package.json
```diff
@@ -8,10 +8,14 @@
   "description": "API for SPARK",
   "scripts": {
     "migrate": "node bin/migrate.js",
-    "start": "node bin/spark.js",
+    "start:api": "node spark-api/bin/spark.js",
+    "start:publish": "node spark-publish/bin/spark-publish.js",
+    "test:api": "mocha spark-api/test/**/*.js",
+    "test:publish": "mocha spark-publish/test/**/*.js",
     "test": "standard && mocha"
```
juliangruber, line 15: should this instead do

```bash
npm-run-all test:*
```

spark-api/lib/ie-contract.js
```diff
@@ -1,5 +1,5 @@
 import { ethers } from 'ethers'
-import { RPC_URL, GLIF_TOKEN } from '../spark-publish/ie-contract-config.js'
+import { RPC_URL, GLIF_TOKEN } from '../../spark-publish/ie-contract-config.js'
```
juliangruber, line 2: what about moving `ie-contract-config.js` to a shared top level folder?

.dockerignore
juliangruber, line 1: It looks like `.dockerignore` files were only removed, and none added. Is this intentional?

spark-publish/test/test.js
```diff
@@ -4,7 +4,7 @@ import { CID } from 'multiformats/cid'
 import pg from 'pg'
 import * as telemetry from '../lib/telemetry.js'
 
-import { assertApproximately } from '../../test/test-helpers.js'
+import { assertApproximately } from '../../spark-api/test/test-helpers.js'
```
juliangruber, line 7: shall we move these to a shared folder as well?

README.md
```diff
@@ -89,13 +89,13 @@ docker run -d --name spark-db \
 Start the API service:
 
 ```bash
-npm start
+npm run start:api
```
juliangruber, line 92: The only workflow downside that I see with this approach is that you can't as easily zoom in into projects any more: If I open my editor on `./spark-api/`, I need to keep my CWD in `.` and remember to run `npm run test:api`. It would be nice if I could `cd` into that folder and just run `npm test`. Is this a concern for you?

What about instead we give every subrepo their own package.json and node_modules? We can still have a parent package.json on the top level, which will shell out to the sub-parent.jsons for running tests etc. I think this matches the lerna experience closer as well. But are there downsides to this approach?

juliangruber (changes requested):
Great progress, Miro! 👏

Dockerfile
bajtos: Great question!

FWIW, the Fly.io project is called `spark`, I think the name `spark-api` is already taken by somebody else. But you are right that all other Fly.io project names contain the `spark-` or `voyager-` prefixes.


I was thinking that maybe we can change the directory names to drop the prefix and rename `spark-api/spark-api` to `spark-api/api`, `spark-api/spark-publish` to `spark-api/publish`, and so on.

However, I also see how using the Fly project name as the directory name makes things easier to understand.

In that light, should we use the names `spark` (for API) and `spark-publish`?

The downside is that we have more to type at dev time, but I like how the script suffix matches the directory (sub-package) name.

```
npm run test:spark-publish
```

The remaining question is how to call the directory with the REST API server - should it be `spark-api` to match what it holds or `spark` to match the Fly project? I prefer `spark-api` as I find it less confusing.

package.json
bajtos: > should this instead do
> 
> ```shell
> npm-run-all test:*
> ```

It's an option too. Let's discuss the difference:

- In my version, we run a single Mocha process that can run tests for all components in parallel.
- In your version, we run multiple Mocha processes, one for each component. I checked the docs and nom-run-all supports both parallel and serial execution, which is great. 

  The downside is that we will call `npm run` several times, which adds overhead and slows down the feedback loop. I did a quick test, and it seems that `npm run` adds 160ms.

```
❯ time (echo ok)                                                                                   
ok
( echo ok; )  0.00s user 0.00s system 39% cpu 0.002 total
❯ time npm run echo

> @filecoin-station/spark-api@0.0.0 echo
> echo ok

ok
npm run echo  0.16s user 0.03s system 105% cpu 0.186 total
```

spark-api/lib/ie-contract.js
bajtos: That's a great idea!

.dockerignore
bajtos: That was a mistake, I forgot to commit the file. Fixed in [c7d99f5](https://github.com/filecoin-station/spark-api/pull/323/commits/c7d99f58ae34e41b67ab5474590acb6bc352f203)

README.md
bajtos: > The only workflow downside that I see with this approach is that you can't as easily zoom in into projects any more: If I open my editor on `./spark-api/`, I need to keep my CWD in `.` and remember to run `npm run test:api`. It would be nice if I could `cd` into that folder and just run `npm test`. Is this a concern for you?

Makes sense. I agree we should at least try to make this workflow easier.

I usually open my editor/IDE in the monorepo root, so this is not a concern to me. 

My thoughts:
- When you run `npm` in the subdirectory and there is no package.json file, npm will walk up the directory tree to the monorepo root and use the package.json file in the root.
- However, `npm test` will set the working directory to the directory containing the `package.json` file, so it will run all tests.
- You can run something like `npx mocha test`, but that's ugly.
- You can also run `npm run test:spark-api`, but that's not ideal either.

We can create a small `package.json` file containing only the scripts:

```json
{
  "type": "module",
  "scripts": {
    "test": "cd ..; npm run test:spark-api"
  }
}
```

Do you see any downsides with that?

> What about instead we give every subrepo their own package.json and node_modules? We can still have a parent package.json on the top level, which will shell out to the sub-parent.jsons for running tests etc. I think this matches the lerna experience closer as well. But are there downsides to this approach?

Yes, this is the Lerna/`npm workspace` setup.

From what I remember, this setup makes many tasks slower:
- `npm install` needs to install multiple copies of the same package when it's used by multiple components.
- Depending on the install strategy, you can end up with all packages installed in the root `node_modules` folder. As a result, the code in `spark-publish` can import dependencies declared in `spark-api` only and we won't notice that.
- We probably need to configure TypeScript with a multi-project setup. The last time I used such a setup in a large monorepo (~2021), it took seconds to minutes for `tsc` just to load the entire tree and decide what needs to be (re)checked.

Anyhow, I am open to giving this a try.

Dockerfile
```diff
@@ -40,5 +38,12 @@ FROM base
...
+# This argument controls what npm script is executed as the CMD.
+ARG SERVICE
+
+# ARGs are not preserved at runtime, we need to store the value
+# as a default value of an ENV var
+ENV SCRIPT="start:${SERVICE}"
+
+CMD npm run ${SCRIPT}
```
bajtos, line 49: @juliangruber do you have any opinion on the best Dockerfile & deployment configuration?

1. In the current proposal, we hard-code the service name at build time, but it can still be changed at run time via the `SCRIPT` env var
2. An alternative is to drop `ARG SERVICE` completely, let all services use the same Docker image, and change `fly.toml` config - remove `[build.args]` section and add `SCRIPT` env var config.
3. Another alternative is to use multi-stage build and add a new stage for each service. 

    ```Dockerfile
    FROM base as spark-api
    COPY --from=build /app /app
    CMD [ "npm", "run", "start:spark-api" ]
    
    FROM base as spark-publish
    COPY --from=build /app /app
    CMD [ "npm", "run", "start:spark-publish" ]
    ```

    We can specify the build target in the `fly.toml` file, see https://fly.io/docs/reference/configuration/#specify-a-multistage-docker-build-target

    ```toml
    [build]
      build-target = "spark-api"
    ```

Personally, I like the second option most. It is a bit simpler than the current `ARG SERVICE` variant.

README.md
bajtos: Documentation on different install strategies for npm workspaces:

https://docs.npmjs.com/cli/v10/commands/npm-ci?v=true#configuration

bajtos: I opened a PR against this feature branch to try out npm workspaces: https://github.com/filecoin-station/spark-api/pull/326

Dockerfile
juliangruber: >  I think the name spark-api is already taken by somebody else

I don't think it is, or I haven't seen it. I called it `spark` in Fly because I assumed at that time it would be the only service we'd create for it :D I'm in favor of renaming it to `spark-api` in Fly.

I would prefer to go with `spark-api/{api,publish}`, I think we can remember that `spark-api/x` maps to `spark-x` on Fly. But the other option is fine for me too!

juliangruber: We can also change this in the future, if we change our mind

package.json
juliangruber: Oh I see, I didn't realize that `mocha` would run all test suites. That's perfect then!

README.md
juliangruber: > Do you see any downsides with that?

The downside would be that if you add a dependency from the subdirectory, it won't be added to the right place.

> Yes, this is the Lerna/`npm workspace` setup.
> From what I remember, this setup makes many tasks slower:
> - `npm install` needs to install multiple copies of the same package when it's used by multiple components.

To me, `npm install` time doesn't matter these days. I don't run it often, and the cache + my internet are fast enough. If it's a significant issue on your setup, or for your workflow, I'm happy to value this more.

> - Depending on the install strategy, you can end up with all packages installed in the root node_modules folder. As a result, the code in spark-publish can import dependencies declared in spark-api only and we won't notice that.

In which situation would all packagers be installed in the root folder? I believe that when you run `cd api && npm install` and there's `api/package.json`, then dependencies _have_ to go to `abi/node_modules/`.

> - We probably need to configure TypeScript with a multi-project setup. The last time I used such a setup in a large monorepo (~2021), it took seconds to minutes for tsc just to load the entire tree and decide what needs to be (re)checked.

Ah right I remember that from our talk. I definitely don't want to introduce TS headaches here.

Dockerfile
juliangruber: Tbh I don't have a preference here as all seems straight forward to me. Please choose what you prefer

README.md
bajtos: > To me, `npm install` time doesn't matter these days. I don't run it often, and the cache + my internet are fast enough. If it's a significant issue on your setup, or for your workflow, I'm happy to value this more.

I kind of resigned on slow npm installs, until I had a chance to experience how fast is `pnpm` and ever since then I am missing its speed. Switching from npm to pnpm is out of scope of this work though.

> > * Depending on the install strategy, you can end up with all packages installed in the root node_modules folder. As a result, the code in spark-publish can import dependencies declared in spark-api only and we won't notice that.
> 
> In which situation would all packagers be installed in the root folder? I believe that when you run `cd api && npm install` and there's `api/package.json`, then dependencies _have_ to go to `abi/node_modules/`.

npm's default configuration is to hoist dependencies. It will always install as many dependencies as possible in the root node_module folder.

> [install-strategy](https://docs.npmjs.com/cli/v10/commands/npm-ci/?v=true#install-strategy)
> 
> Default: "hoisted"
> Type: "hoisted", "nested", "shallow", or "linked"
> Sets the strategy for installing packages in node_modules. hoisted (default): Install non-duplicated in top-level, and duplicated as necessary within directory structure. nested: (formerly --legacy-bundling) install in place, no hoisting. shallow (formerly --global-style) only install direct deps at top-level. linked: (experimental) install in node_modules/.store, link in place, unhoisted.

> > * We probably need to configure TypeScript with a multi-project setup. The last time I used such a setup in a large monorepo (~2021), it took seconds to minutes for tsc just to load the entire tree and decide what needs to be (re)checked.
> 
> Ah right I remember that from our talk. I definitely don't want to introduce TS headaches here.

We also don't have TypeScript checks configured in this repository yet. I am willing to take a risk and not worry about this until I start working on adding `tsc`.

I think we should be fine as long as our monorepo stays small (less than 10 sub-packages, less than 100 source files).

bajtos:
Next steps:

- Finish https://github.com/filecoin-station/spark-api/pull/326 and merge it into this feature branch
- Rename `spark-{name}` directories to `{name}` per https://github.com/filecoin-station/spark-api/pull/323#discussion_r1618148171
- Move `ie-contract-config.js` to a shared top-level folder per https://github.com/filecoin-station/spark-api/pull/323#discussion_r1618149183
  - Proposed name: `/common/ie-contract-config.js`
- Move test helpers to a shared top-level folder per https://github.com/filecoin-station/spark-api/pull/323/files/a668dae332d188b8959943442267ada481eacea3#diff-25bf6f781fa9084fcfc6e5671da39549a4a15bab23b166f7f544d344fe43ab4a
  - Proposed name: `/test-helpers/assert.js`
- Simplify Dockerfile to use only `ENV`-based config per https://github.com/filecoin-station/spark-api/pull/323#pullrequestreview-2084511396

PR #326, opened 2024-05-29
feat: npm workspaces

bajtos:
Rework the monorepo layout to use [npm workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) and per-component `package.json` files.

TODO:
- [x] linters
- [x] CI build
- [x] Dockerfile(s)
- [x] Dependabot

Documentation on different install strategies for npm workspaces:
https://docs.npmjs.com/cli/v10/commands/npm-ci?v=true#configuration


.github/workflows/ci.yml
```diff
@@ -51,14 +51,25 @@ jobs:
...
+
+  lint-all:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+      - run: npm ci
+      - run: npm run lint
```
bajtos, line 72: I need to run `standard` for all files in the monorepo, including those that are not in any sub-package, e.g. `migrations/index.js`

spark-api/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "name": "@filecoin-station/spark-api",
+  "private": true,
+  "version": "0.0.0",
+  "license": "MIT",
+  "repository": "filecoin-station/spark-api",
+  "type": "module",
+  "description": "API for SPARK",
+  "scripts": {
+    "start": "node bin/spark.js",
+    "test": "mocha"
```
bajtos, line 11: In the main branch, one could run `(cd spark-publish && npm test)` to execute both the linter and the tests for a sub-package.

Is this something you want to preserve in the new monorepo setup?

In my proposal, you always have to run the linter from the monorepo root, because that's the place where `package.json` tells `standard` about Mocha.

.github/workflows/ci.yml
juliangruber: Could we turn `migrations` into a sub-package too?

package.json
```diff
@@ -1,46 +1,20 @@
...
+  ],
   "scripts": {
     "migrate": "node bin/migrate.js",
-    "start:api": "node spark-api/bin/spark.js",
-    "start:publish": "node spark-publish/bin/spark-publish.js",
-    "test:api": "mocha spark-api/test/**/*.js",
-    "test:publish": "mocha spark-publish/test/**/*.js",
-    "test": "standard && mocha"
+    "lint": "standard",
+    "test": "npm run lint && npm test --workspaces"
```
juliangruber, line 14: clean 👍

spark-api/package.json
juliangruber: I would prefer indeed if both would work: linting from the parent and the subpackage. Could we create a standard config that re-exports the parent config?

juliangruber (commented):
This setup seems to work well for me - but I haven't stress tested it yet

.github/workflows/ci.yml
bajtos: > Could we turn `migrations` into a sub-package too?

> I would prefer indeed if both would work: linting from the parent and the subpackage. Could we create a standard config that re-exports the parent config?

I am not a fan of this. It adds a lot of boilerplate. We will need to add package.json to `migrations`, add scripts like `lint`, add `standard` as a dependency, and so on.

Having written that, I also want to get this monorepo setup finished ASAP and choose the setup that works for both of us, so I am going to take the direction you suggested.

bajtos: > Could we turn migrations into a sub-package too?

Hmm, we would have to turn `bin` and `scripts` into sub-packages too. I definitely don't like that.

I am going to explore a different setup, where the monorepo root is a package on its own, and then some of the subdirectories are packages (workspaces) too.

spark-api/package.json
bajtos: Done in https://github.com/filecoin-station/spark-api/pull/326/commits/daa26500505d69998ac67522d8b53eaa72ab5065

.github/workflows/ci.yml
bajtos: Done in https://github.com/filecoin-station/spark-api/pull/326/commits/daa26500505d69998ac67522d8b53eaa72ab5065

bajtos:
Dependabot's support for npm workspaces is a mixed bag. I guess we have to bite the bullet and hope all will work well.

- https://github.com/dependabot/dependabot-core/issues/5226
- https://github.com/dependabot/dependabot-core/issues/4993#issuecomment-1739296001
- https://github.com/dependabot/dependabot-core/issues/6346
- https://github.com/aws-powertools/powertools-lambda-typescript/pull/1917
- https://github.com/dependabot/dependabot-core/issues/6270
- https://github.com/dependabot/dependabot-core/issues/7441

.github/workflows/ci.yml
bajtos: I am keeping the `lint-all` step so that we can lint all workspaces in one go, and in parallel with running the tests.

The alternative is to lint `spark-api` and `spark-publish` as part of `build-*` workflows, and then add another workflow to run only files that are not part of any child workspace (`bin`, `migrations`, `scripts`). I am not sure how easy that would be to do, though, so I prefer the current proposal.

spark-api/package.json
bajtos: > Could we create a standard config that re-exports the parent config?

I could not find any quick way how to do that, Standard seems to accept config only via CLI args or in the closest `package.json` file.

The alternative is to eject from standard to eslint, but that's much more work than I am willing to do now.

We could also remove the env config from package.json and modify every file to start with the following line:

```
/* eslint-env mocha */
```

It would be cleaner because non-test files won't have access to `describe` & `it` globals, but it also feels like too much boilerplate.

bajtos: I ended up duplicating the `"standard":{"env":...}}` section in each `package.json`.

PR #339, opened 2024-06-05
feat: log IP Group and location info to InfluxDB

PatrickNercessian:


api/lib/network-management.js
```diff
@@ -103,3 +105,33 @@ export const mapRequestToSubnet = (req) => {
...
+    INSERT INTO network_info_update_history (station_id, updated_at)
+    VALUES ($1, NOW())
+    ON CONFLICT (station_id) DO UPDATE
+    SET updated_at = NOW()
+    WHERE network_info.updated_at < NOW() - INTERVAL '10 minutes'
+    RETURNING station_id
+  `, [stationId])
+
+  // Don't record to InfluxDB if we didn't update the row
+  if (rows.length === 0) return
```
PatrickNercessian, line 121: Thoughts on this as a method of avoiding excessive InfluxDB writes?

api/lib/telemetry.js
```diff
@@ -0,0 +1,24 @@
...
+  writeClient.flush().catch(console.error)
+}, 10_000).unref()
+
+export const record = (name, fn) => {
+  const point = new Point(name)
+  fn(point)
+  writeClient.writePoint(point)
+}
+
+export const close = () => writeClient.close()
```
PatrickNercessian, line 24: Are we okay with duplicating this telemetry.js file?

PatrickNercessian:
Seems the build-api step is failing because it's missing the GLIF_TOKEN env variable. Not sure how my changes would have caused this

PatrickNercessian:
Link to roadmap issue: https://github.com/filecoin-station/roadmap/issues/112

api/lib/network-management.js
juliangruber: I don't think we need 10 minutes precision, recording once per day is good enough. And with 1d precision, I think we don't need to compare to only record when the network changed, we can just record always. Could you simplify the setup with this? I think we only need to track whether we already recorded for this station id in this day.

Since this query will be performed for every measurement recorded, performance is crucial.

In pseudo code, the implementation could be somewhat like this:

```js
const stationIDsSeen = []
setInterval(() => { stationIDsSeen.length = 0 }, ONE_DAY)
// ...
if (!stationIDsSeen.includes(stationID)) {
  stationIDsSeen.push(stationID)
  record()
}
```

The list of Station IDs should be small enough to fit in memory easily. The only downside of using in-memory is that after deploy every station will get recorded again. Therefore I think we still need to use the DB for this

juliangruber (changes requested):

api/lib/network-management.js
PatrickNercessian: All makes sense to me, but your last sentence

> Therefore I think we still need to use the DB for this

Do you mean we you think we shouldn't use the in-memory list and instead go with the current change using a new SQL table, but change it to 24h?

api/lib/telemetry.js
PatrickNercessian: Ah didn't even see this, thanks!

api/lib/network-management.js
juliangruber: I was wondering if we could just use an in-memory structure - the participants list is small enough. But, the downside is that we will record all participants after a restart, so I think it's not an option.

I think the best solution is to use an SQL table, but instead of comparing dates on INSERT (expensive), periodically empty the table

api/package.json
```diff
@@ -20,7 +20,12 @@
   "dependencies": {
     "@filecoin-station/spark-impact-evaluator": "^1.1.1",
     "@glif/filecoin-address": "^3.0.5",
+<<<<<<< feat-log-network-info
```
juliangruber, line 23: merge conflicts committed

PatrickNercessian: Weird, I used GitHub's editor for once and removed it, but maybe I needed to save the page or something before marking as resolved.

Addressed in https://github.com/filecoin-station/spark-api/pull/339/commits/d169b6d253000391c888ed9ff090eea6d3ae0773

api/test/inet-grouping.test.js
```diff
@@ -126,3 +191,16 @@ const createRequestFromAddress = (remoteAddress, headers) => {
...
+    INSERT INTO network_info_update_history (day, station_id)
+    VALUES ($1, $2)
+    ON CONFLICT (day, station_id) DO NOTHING
+  `, [day, stationId])
+}
+
+const getCurrentDate = async () => {
+  const { rows: [{ today }] } = await pgClient.query('SELECT now()::DATE::TEXT as today')
+  return today
+}
```
bajtos, line 206: I have seen this code in other tests files, what do you think about extracting it into a shared test helper and create a new file in the [/test-helpers](https://github.com/filecoin-station/spark-api/tree/main/test-helpers) directory?

api/lib/inet-grouping.js
```diff
@@ -103,3 +103,42 @@ export const mapRequestToSubnet = (req) => {
...
+  `, [stationId])
+
+  // Don't record to InfluxDB or delete the row if we didn't write the row
+  if (rows.length === 0) return
+
+  // Empty the table of every day before today
+  await pgClient.query(`
+    DELETE FROM network_info_update_history
+    WHERE day < NOW() - INTERVAL '1 day'
+  `)
```
bajtos, line 129: IIUC, this implements a locking scheme that supports multiple Node.js processes, using SQL as the shared state. Great idea! 👏🏻 

It makes me wonder if the design based on `(day, station_id)` is the most efficient one and whether it's worth optimising this part a bit more.

Alternative to consider - I am not sure how much better it is, just sharing it here for discussion.

A benefit of my proposalis that it makes it super easy to tweak the "once per day" interval in the future. We may not ever need that; I am mentioning it just for completeness.

(1) Change the table to have `station_id` and `last_updated_at` columns, using `station_id` as the PK. 

(2) Change the implementation here as follows:

```js
const { rows } = await pgClient.query(`
  UPDATE network_info_update_history 
  SET last_updated_at = NOW() 
  WHERE station_id = $1 AND last_updated_at < NOW() - INTERVAL '1 day'
  RETURNING 1 as needs_update
`, [stationId])

if (rows.length === 0) return

await pgClient.query(`
  DELETE FROM network_info_update_history
  WHERE last_updated_at < NOW() - INTERVAL '1 day'
`)
```

(3)
Create an index over `last_updated_at` or perhaps `(last_updated_at, station_id)`. I think the second form should support both queries above, but this needs to be verified using `EXPLAIN` or `EXPLAIN ANALYZE`.

api/lib/inet-grouping.js
```diff
@@ -103,3 +103,42 @@ export const mapRequestToSubnet = (req) => {
...
 }
+
+/**
+ * @param {import('pg').Client} pgClient
+ * @param {import('node:http').IncomingMessage} req
+ * @param {string} stationId
+ * @param {string} inetGroup
+ * @param {function} recordTelemetryFn
+ */
+export const logNetworkInfo = async (pgClient, headers, stationId, inetGroup, recordTelemetryFn) => {
```
bajtos, line 114: I agree it's better to keep the name `inet-grouping.js`. However, I don't think this new function `logNetworkInfo` belongs to a file with such a name.

Can we put this function to a new file please? For example `network-info-logger.js`.

@juliangruber what's your opinion?

bajtos (changes requested):

api/lib/inet-grouping.js
juliangruber: Since it's only used by one file, I'd also like to inline it in that file. Otherwise your proposal works for me too

api/lib/inet-grouping.js
```diff
@@ -103,3 +103,42 @@ export const mapRequestToSubnet = (req) => {
...
+    point.stringField('inet_group', inetGroup)
+
+    point.stringField('cf-ipcity', headers['cf-ipcity'])
+    point.stringField('cf-ipcountry', headers['cf-ipcountry'])
+    point.stringField('cf-ipcontinent', headers['cf-ipcontinent'])
+    point.stringField('cf-iplongitude', headers['cf-iplongitude'])
+    point.stringField('cf-iplatitude', headers['cf-iplatitude'])
+    point.stringField('cf-region', headers['cf-region'])
+    point.stringField('cf-region-code', headers['cf-region-code'])
+    point.stringField('cf-timezone', headers['cf-timezone'])
```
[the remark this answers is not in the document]
bajtos: Quoting from the [Architectural Proposal](https://docs.google.com/document/d/18L0PdxV5nAYuTbeyMNkOCRITM32PFyQtYZNeGbQVOPY/edit?pli=1):

> - To preserve users’ privacy, we should not store their IP address or overly granular location data using decentralized/open-source methods, nor display them in our dashboard. Instead, we can use CloudFlare’s categorical location information (country, city, region).
> - We may consider not storing any location information at all tying a certain wallet address to any location, no matter how general.

common/telemetry.js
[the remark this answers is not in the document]
bajtos: Sorry, I wanted to say "a dedicated bucket", not "buffer". 🙈

api/lib/network-management.js
juliangruber: sorry, which part was too expensive?

PatrickNercessian: Sorry should have clarified, from your comment above:

> I think the best solution is to use an SQL table, but instead of comparing dates on INSERT (expensive), periodically empty the table

bajtos: @PatrickNercessian I discussed this with @juliangruber today. We decided it's best to radically simplify the first version and then iteratively improve it as needed.

Let's drop the entire `network_info_update_history` table & related code and keep the state in memory.

We consider the downside of having duplicate data points (caused by the in-memory state not shared by the four processes/service instances) to be acceptable.

We did a bit of napkin math: in the worst case, we have ~10 deployments per day, which means 40 duplicate submissions of the geolocation data. With 20k unique station instances, that means ~800k data points daily. We think that's not unreasonable, considering how many other different data points we are writing from other parts of the Station & Spark systems.

If this becomes a problem, we can add a caching layer then.

Could you please rework the pull request accordingly?

Also, remember to use a different bucket for these location data points; we will want to set a very short data retention period - something like 3 days.

PatrickNercessian: Are we sure? The only listed downside to the existing PR approach is that we can't easily modify the fact that it updates every 24h.

The in-memory state does not give us this functionality either, and adds the unnecessary duplicate submissions (even if they're not a real problem at the moment), and, to me, it does not seem obvious that it is much simpler.

Actually, as I'm thinking about it, I guess we could have the "updated_at" timestamp in-memory as well, and use JS logic rather than SQL logic to decide whether to update? Is this what you meant?

Just want to double check, if you're certain about the in-memory state I can go ahead and start implementation!

bajtos: We are concerned about the performance impact of making another SQL query every time we store a new measurement.

The idea is to remove all SQL from this pull request and keep the update history in memory.

If this becomes a problem, we would like to explore more options:
- share the update history via Redis
- query InfluxDB to find the last update time
- add in-memory cache for those InfluxDB requests
- store the update history in SQL as you proposed here
- maybe something else

The point is to ship the code collecting location data first, observe how it works in production, and only then fix what needs fixing.

PatrickNercessian: Ok! So were you thinking we do an "updated_at" timestamp (or an "last_update_day" date) in-memory, with JS logic?

juliangruber: I think something like this could still work, as posted in the beginning of the thread:

```
const stationIDsSeen = []
setInterval(() => { stationIDsSeen.length = 0 }, ONE_DAY)
// ...
if (!stationIDsSeen.includes(stationID)) {
  stationIDsSeen.push(stationID)
  record()
}
```

api/lib/inet-grouping.js
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-api/pull/339/commits/66b9d3e08937267824c36c61f708c182230d187e. I did 'network-info-logger.js' to avoid a weird const `stationIdsSeen` at the top of index.js, and for a good place to put unit tests.

Let me know if either of you prefer it in index.js

api/test/inet-grouping.test.js
PatrickNercessian: Makes sense! Due to more recent discussion, this is not needed anymore due to in-memory structure instead though

common/telemetry.js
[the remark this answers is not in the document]
PatrickNercessian: Used `spark-api` bucket, with the point name being network-info: https://github.com/filecoin-station/spark-api/pull/339/commits/66b9d3e08937267824c36c61f708c182230d187e

I assume we need to create this bucket in InfluxDB though

api/lib/inet-grouping.js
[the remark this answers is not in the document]
PatrickNercessian: Removed station_id and inet_group in https://github.com/filecoin-station/spark-api/pull/339/commits/66b9d3e08937267824c36c61f708c182230d187e

api/lib/network-management.js
PatrickNercessian: Got it! Addressed in https://github.com/filecoin-station/spark-api/pull/339/commits/66b9d3e08937267824c36c61f708c182230d187e

publish/test/test.js
```diff
@@ -2,13 +2,13 @@ import { publish } from '../index.js'
...
 import pg from 'pg'
-import * as telemetry from '../lib/telemetry.js'
+import * as telemetry from '../../common/telemetry.js'
 
 import { assertApproximately } from '../../test-helpers/assert.js'
 
 const { DATABASE_URL } = process.env
 
-after(telemetry.close)
+after(async () => { await telemetry.publishWriteClient.close() })
```
bajtos, line 11: IIRC, we are calling `close` to prevent the write client from blocking the process from exiting.

Ideally, tests should not write any telemetry, they should mock the `recordTelemetryFn`, as we already do in other tests.

Alternatively, the `telemetry` module should implement a `close` function to close all writers.

For example (may not work out of the box):

```js
// usage here - no change
after(telemetry.close)

// implementation in common/telemetry
export const close = async () => {
  await Promise.all([
    publishWriteClient.close(), 
    apiWriterClient.close(), 
    //etc.
  ])
}
```

api/test/network-info-logger.test.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+  const headers3 = {}
+  for (const key in headers1) {
+    headers3[key] = headers1[key].slice(0, -1) + '3'
+  }
+
+  it('should record new network info if not present for the day', async () => {
+    await logNetworkInfo(headers1, 'station-id1', recordTelemetry)
+    await logNetworkInfo(headers2, 'station-id2', recordTelemetry)
+    await logNetworkInfo(headers3, 'station-id1', recordTelemetry)
```
bajtos, line 39: Let's add a comment to make it clear that the third call is the place where we test the condition _"if not present for the day"_.

```suggestion
    // another request from a Station ID we have already seen today
    await logNetworkInfo(headers3, 'station-id1', recordTelemetry)
```

api/test/network-info-logger.test.js
```diff
@@ -0,0 +1,56 @@
...
+      expectedFields1[key] = `"${headers1[key]}"`
+    }
+    const expectedFields2 = {}
+    for (const key in headers2) {
+      expectedFields2[key] = `"${headers2[key]}"`
+    }
+
+    assert.strictEqual(telemetry.length, 2)
+    assert.strictEqual(telemetry[0].name, 'network-info')
+    assert.deepStrictEqual(telemetry[0].fields, expectedFields1)
```
bajtos, line 52: These assertions work well on the happy path. On failure, they will produce error messages that does not provide information needed to understand what went wrong.

For example, if `telemetry.length` is `1` instead of `2`, we cannot tell what exactly happened.

Something like this would produce better errors IMO.

```js
assert.deepStrictEqual(
  telemetry.map(p => { _point: p.name, ...p.fields }), 
  [
    { _point: 'network-info', ...expectedFields1 },
    { _point: 'network-info', ...expectedFields2 },
  ]
)
```

(Assertions usually sort object keys lexicographically when presenting a diff. By prefixing the point prop name with `_`, it should always be at the top.)

publish/test/test.js
PatrickNercessian: This closing logic was already present I think, this change is just to bring it in line with the new telemetry file.

If we want to change this, maybe it's better scoped in a different PR?

api/lib/network-info-logger.js
```diff
@@ -0,0 +1,28 @@
+const stationIdsSeen = new Set()
+setInterval(() => { stationIdsSeen.clear() }, 1000 * 60 * 60 * 24) // clear every day
+
+/**
+ * @param {import('node:http').IncomingMessage} req
+ * @param {string} stationId
+ * @param {string} inetGroup
+ * @param {function} recordTelemetryFn
```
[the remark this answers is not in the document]
PatrickNercessian: Would you want a new file like in spark-evaluate for the typings? Or just have it in telemetry.js or network-info-logger.js?

api/index.js
```diff
@@ -78,6 +80,7 @@ const createMeasurement = async (req, res, client) => {
   assert(measurement.stationId.match(/^[0-9a-fA-F]{88}$/), 400, 'Invalid Station ID')
 
   const inetGroup = await mapRequestToInetGroup(client, req)
+  logNetworkInfo(req.headers, measurement.stationId, recordApiTelemetry)
```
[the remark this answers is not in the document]
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-api/pull/339/commits/ae602bb01cb68378b49bb7bb3f80c7439e61f93c!

common/telemetry.js
```diff
@@ -0,0 +1,40 @@
...
+  'Filecoin Station', // org
+  'spark-publish', // bucket
+  'ns' // precision
+)
+
+const apiWriteClient = influx.getWriteApi(
+  'Filecoin Station', // org
+  'spark-api', // bucket
+  'ns' // precision
+)
```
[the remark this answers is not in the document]
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-api/pull/339/commits/ae602bb01cb68378b49bb7bb3f80c7439e61f93c!

api/test/network-info-logger.test.js
PatrickNercessian: I like it better too, addressed in https://github.com/filecoin-station/spark-api/pull/339/commits/3b7649985da67301eeda32e92c1c43af44c77d73

api/lib/network-info-logger.js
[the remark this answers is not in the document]
bajtos: I prefer to add `common/typings.d.ts` file for consistency with the rest of our projects.

api/lib/inet-grouping.js
bajtos: We agreed to use an in-memory structure, so this thread is no longer relevant.

common/telemetry.js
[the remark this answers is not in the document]
bajtos: Let's continue the discussion in https://github.com/filecoin-station/spark-api/pull/339#discussion_r1652183787.

api/index.js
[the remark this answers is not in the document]
bajtos: Great!

Quoting from an earlier comment:

> Used spark-api bucket, with the point name being network-info.
> I assume we need to create this bucket in InfluxDB though

I created the bucket `spark-network-info`.

Screenshot 2024-06-26 at 08 06 28

publish/test/test.js
bajtos: I see your point.

The problem I am trying to prevent is that with your changes in place, the test suite does not exit cleanly, because there are background active writers.

Before your change, when executing the tests in `publish/test/test.js`, only one writer was created and this writer was closed by this `after` line. 

With your changes in place, executing this test file creates more than on writer, but we close only one of them. As a result, the test suite never exits.

I confirmed this by running the following command:

```
npm t -w api -- -i -g "getRoundStartEpoch|startRoundTracker"
```

We want to rework this test file anyway; it should not write any telemetry, as I mentioned before. Essentially, we want to port https://github.com/filecoin-station/voyager-api/pull/22 to this repo.

I am going to make two smaller changes: fix this test file to use a mocked telemetry recorder and introduce `common/typings.d.ts` file as discussed in https://github.com/filecoin-station/spark-api/pull/339#discussion_r1652173812

bajtos: See https://github.com/filecoin-station/spark-api/pull/354

In the end, I did not need `common/typings.d.ts`, so I am leaving that change for you to make in this pull request.

bajtos: I think we can remove `telemetry.close()` altogether now, including the implementation.

api/lib/network-info-logger.js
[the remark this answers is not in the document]
bajtos: In other places, we use the variable/argument name `recordTelemetry`. Please do the same here for consistency.

publish/test/test.js
bajtos:
```suggestion
```

api/test/network-info-logger.test.js
```diff
@@ -0,0 +1,59 @@
+import assert from 'node:assert'
+import { Point } from '../../common/telemetry.js'
+import { logNetworkInfo } from '../lib/network-info-logger.js'
+
+const telemetry = []
+const recordTelemetry = (measurementName, fn) => {
+  const point = new Point(measurementName)
+  fn(point)
+  telemetry.push(point)
+}
```
bajtos, line 10: Please use the test helper I introduced in #354 and made available to API tests in #355.

Also, the list of recorded telemetry points should be reset between the tests, so it's best to move the initialisation into each test.

```js
it('should record new network info if not present for the day', async () => {
  const {telemetry, recordTelemetry} = createTelemetryRecorderStub()
  await logNetworkInfo(headers1, 'station-id1', recordTelemetry)
  // etc.
})

**Review by bajtos — CHANGES_REQUESTED** [review]
Almost there! 🤩 

Sorry about the merge conflicts I created by my two PRs and left for you to resolve. 🙈 

Besides the comment below, there are a couple of unresolved conversations above; please look at them, too.

File: api/lib/inet-grouping.js (hunk 6 above)
Comment (bajtos) [reply api/lib/inet-grouping.js]: I think this is good enough for now 👍🏻

File: publish/bin/publish-batch.js [hunk 15]
```diff
@@ -60,13 +60,13 @@ try {
     client,
     web3Storage,
     ieContract,
-    recordTelemetry,
+    recordPublishTelemetry,
```
bajtos, line 63: This rename doesn't look right to me. The `publish` function expects `recordTelemetry` argument. 


Alternatively, we could rename the argument. I would find it a bit repetitive to have a `publish` function with `recordPublishTelemetry` param. I prefer the current version more.

```suggestion
    recordTelemetry: recordPublishTelemetry,
```

test-helpers/platform-test-helpers.js
```diff
@@ -1,7 +1,7 @@
 // This file is shared with voyager-api/voyager-publish
 // Helpers in this file must not have anything project-specific like measurement fields
 
-import { Point } from '../publish/lib/telemetry.js'
+import { Point } from '../common/telemetry.js'
```
bajtos, line 4: Perhaps we should import the Point directly from InfluxDB SDK to avoid calling the code that initialises the writers?

api/lib/network-info-logger.js
```diff
@@ -0,0 +1,27 @@
+const stationIdsSeen = new Set()
+setInterval(() => { stationIdsSeen.clear() }, 1000 * 60 * 60 * 24) // clear every day
```
bajtos, line 2: I think this blocks the event loop from exiting when running the tests. 

```
npm t -w api -- -i -g "getRoundStartEpoch|startRoundTracker"
```

Possible solutions that come to my mind:
1. Allow the tests to clear this interval.
2. Initialise `setInterval` in a file not loaded by the tests, e.g. a bin file.

I like the second option more. The problem with the first option is that this file is kind of deep in the tree, and it's not clear which tests are expected to clear the interval.

My proposal:
- In this file, export a function to clear the stations seen
- Setup the interval somewhere else.

This also enables us to test whether the cleaning function works as intended.

bajtos (changes requested):

api/lib/network-info-logger.js
PatrickNercessian: When you say a bin file, do you mean a new one? Doesn't that need to be run directly as part of a node/npm run command?

Or should we instead include this function call in the bin/spark.js file?

PatrickNercessian: Went with the assumption of bin/spark.js, addressed in https://github.com/filecoin-station/spark-api/pull/339/commits/6332977b75dd30bd098b3c751b5efbefb59a2d0e

test-helpers/platform-test-helpers.js
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-api/pull/339/commits/6332977b75dd30bd098b3c751b5efbefb59a2d0e

publish/bin/publish-batch.js
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-api/pull/339/commits/6332977b75dd30bd098b3c751b5efbefb59a2d0e

bajtos (approved):
LGTM 👍🏻 

bajtos:
I created https://github.com/filecoin-station/spark-api/pull/360 to run the test on CI under my account, all passed 👍🏻

PR #341, opened 2024-06-10
Use rpc fallback provider

juliangruber:


bajtos (approved):
Let's give it a try and observe how it works in production.

PR #354, opened 2024-06-26
test(publish): use mocked telemetry recorder

bajtos:
This commit partially ports https://github.com/filecoin-station/voyager-api/pull/22 to this repository:

- Introduce test-helpers and platform-test-helpers
- Rework the test file to use these new helpers

### DIFF
--- a/publish/test/platform-test-helpers.js
+++ b/publish/test/platform-test-helpers.js
@@ -1,0 +1,32 @@
+// This file is shared with voyager-api/voyager-publish
+// Helpers in this file must not have anything project-specific like measurement fields
+
+import { Point } from '../lib/telemetry.js'
+
+export const { DATABASE_URL } = process.env
+
+export const logger = {
+  log () {},
+  error (...args) {
+    console.error(...args)
+  }
+}
+
+export const createTelemetryRecorderStub = () => {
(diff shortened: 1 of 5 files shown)


publish/test/platform-test-helpers.js
bajtos, line 1: I think these helpers are generic enough to be shared by both `publish` and `api` projects. Maybe we should move them to the shared package `/test-helpers`. WDYT, @juliangruber?

bajtos:

publish/test/platform-test-helpers.js
juliangruber: sg, what about `common/test`?

juliangruber (commented):
lgtm

publish/test/platform-test-helpers.js
bajtos: I would expect that `common/test` contains tests for the code in `common`.

If we ever write more complex test helpers, we may want to write tests for these helpers, too. Such tests would go to `/test-helpers/test`.

juliangruber: Hmm maybe I'm confused about the concept of `common` then. I thought either this would contain all shared code, or that it should be removed and each shared lib becomes its own package.

PR #361, opened 2024-06-28
feat: process.env.INFLUXDB_TOKEN

bajtos:
Move the configuration of the writer token from plain-text code to Fly.io secrets.

### DIFF
--- a/common/telemetry.js
+++ b/common/telemetry.js
@@ -1,10 +1,18 @@
 import { InfluxDB, Point } from '@influxdata/influxdb-client'
 
+const {
+  INFLUXDB_TOKEN
+} = process.env
+
+if (!INFLUXDB_TOKEN) {
+  console.warn('Warning: INFLUXDB_TOKEN was not provided by the environment')
+}
+
 const influx = new InfluxDB({
   url: 'https://eu-central-1-1.aws.cloud2.influxdata.com',
-  // spark-publish-write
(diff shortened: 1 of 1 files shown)


publish/lib/telemetry.js
```diff
@@ -1,9 +1,16 @@
...
+} = process.env
+
+if (!INFLUXDB_TOKEN) {
+  console.warn('Warning: INFLUXDB_TOKEN was not provided by the environment')
+}
+
 const influx = new InfluxDB({
   url: 'https://eu-central-1-1.aws.cloud2.influxdata.com',
-  // spark-publish-write
-  token: 'Zkqa_s7mI0W_WKI6DUmu-iRnQkCvwNaQfbPK_zT7I6iYYaC2C1kokdlhO2jb4tjRcAJQHQXAGnrdD3vqlMZ63g=='
```
bajtos, line 6: @juliangruber could you please invalidate (delete) this token in InfluxDB? I could not find it in my account.

bajtos:

I needed to switch to a different token to allow spark-api to write to the spark-network-info bucket (see https://github.com/filecoin-station/spark-api/pull/339) and I figured out this is a good time to convert the token to a secret.

publish/lib/telemetry.js
juliangruber: deleted 👍

PR #385, opened 2024-08-02
api: Add fixed # tasks per round. Closes space-meridian/roadmap#82

juliangruber:
Closes space-meridian/roadmap#82
Depends on https://github.com/filecoin-station/spark-stats/issues/214

Alternatives considered:
- <del>Base task count on measurement count from previous round. This data doesn't exist in spark-stats yet. It should lead to better results though, than the currently used station-id count. @bajtos wdyt? Shall we give this a try and see if it behaves well enough?</del> We went with this alternative

Questions:
- <del>Do we need to test this code path?</del> Yes

TODO
- [x] Track in Graphana
  - [x] measurements target for round
  - [x] measurements count for round
  - [x] total desired measurement count
  - [x] actual measurement count
- [x] Write tests for scaling


juliangruber:
```
+   maxTasksPerNode: 2,
-   maxTasksPerNode: 15,
```

Using live data, this will decrease the job count by a lot. We might want to update the target job volume

api/lib/round-tracker.js
```diff
@@ -11,8 +11,12 @@ import { createMeridianContract } from './ie-contract.js'
...
 export const TASKS_PER_ROUND = 1000
 
-// How many tasks is each SPARK checker node expected to complete every round (at most).
-export const MAX_TASKS_PER_NODE = 15
+// Tweak this to control the network's overall task count.
+export const MAX_TASKS_EXECUTED_PER_ROUND = 500_000
+
+// Baseline value for how many tasks each SPARK checker node is expected to complete every round (at most).
+// The actual value will be set dynamically based on MAX_TASKS_EXECUTED_PER_ROUND and the number of active nodes.
+export const BASELINE_MAX_TASKS_PER_NODE = 15
```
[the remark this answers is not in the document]
bajtos: See https://spacemeridian.grafana.net/d/c4bd3c80-9360-4ea3-8b75-c5ecea907659/spark-internal-dashboard?from=now-2d&to=now&timezone=browser&tab=queries

Screenshot 2024-08-02 at 14 13 13

api/lib/round-tracker.js
```diff
@@ -245,10 +252,30 @@ export async function maybeCreateSparkRound (pgClient, {
...
     roundStartEpoch,
-    MAX_TASKS_PER_NODE
+    BASELINE_MAX_TASKS_PER_NODE
   ])
 
   if (rowCount) {
+    if (useDynamicTaskCount) {
+      try {
+        const res = await fetch('https://stats.filspark.com/stations/daily')
+        const [{ station_id_count: stationIdCount }] = await res.json()
```
bajtos, line 262: I don't think the number of stations per day is trustworthy data right now. We know there are three participants that are churning their Station IDs, adding 900k unique station IDs per day, where most of their station IDs contribute less than 10 measurements only.

> Base task count on measurement count from previous round. This data doesn't exist in spark-stats yet. It should lead to better results though, than the currently used station-id count. @bajtos wdyt? Shall we give this a try and see if it behaves well enough?

Yes, I agree it will be better to use the number of measurements instead of the number of stations.

But then, how do we convert the "desired number of measurements" to "tasks per node" when we don't know how many nodes will submit measurements in this round? We will never know that number, we have to estimate it based on the historical data.

I think it will be great to have a stateful algorithm that's watching how the number of measurements per round evolves from round to round.

Let's say we want to perform MD measurements _per day_. There are 24h/day * 60 minutes/hour / 20 minutes/round = 72 rounds per day. That means we want to record MD/72 measurements per round _averaged over the entire day_.

The logic for deciding on the number of measurements per node for the current round could work like this:

1. Check how many measurements were collected today = T
2. Calculate how many rounds are remaining today = 72 - roundsCompletedToday = RR
3. Now we can calculate the desired "tasks per node" as `(MD - MT) / RR`

The downside is that if the network misbehaves and records way too many measurements early during the day, we will have to perform significantly fewer measurements in the evening. We could possibly stop the network completely if it already exhausted the quote of how many measurements are allowed per day.

We are already storing stats about the number of measurements collected.

- https://stats.filspark.com/retrieval-success-rate returns the total number of all & accepted measurements
- https://stats.filspark.com/miners/retrieval-success-rate/summary returns all & accepted measurement counts with per-miner granularity.

The responses with today's data are cached by Cloudflare for 10 minutes - you may want to disable the cache in your requests.

api/lib/round-tracker.js
```diff
@@ -245,10 +252,30 @@ export async function maybeCreateSparkRound (pgClient, {
...
+        if (stationIdCount === 0) {
+          throw new Error('No active stations found')
+        }
+        await pgClient.query(`
+          UPDATE spark_rounds
+          SET max_tasks_per_node = $1
+          WHERE id = $2
```
bajtos, line 269: I don't like this two-step process where you update the value after it was written to the DB. There is a race condition - if a spark checker requests the current rounds data after you inserted the row but before your updated `max_tasks_per_node`, Cloudflare will store the incorrect response in the cache and serve that incorrect response to other clients (checker nodes) running in the same region.

The fact that there is an HTTP request between these two SQL queries makes that window even wider.

Let's move the code calculating `max_tasks_per_node` to the top, so that the INSERT statement can set the correct `max_task_per_node` value.

bajtos (changes requested):
Great start!

> Do we need to test this code path?

Yes, please. Otherwise, how can we verify that the algorithm works as intended?

api/lib/round-tracker.js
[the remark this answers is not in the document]
bajtos: > This is also influenced by how many measurements we reject because of DUP_INET_GROUP, TASK_WRONG_NODE, TASK_NOT_IN_ROUND and TOO_MANY_TASKS checks, which are difficult to estimate.

We want committees to have at least 30 measurements _after_ discarding the measurements rejected by the checks above.

We don't need all committees to have at least 30 measurements; I think it's ok to discard up to 1-5% of committees because they are too small.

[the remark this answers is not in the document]
bajtos: We can tweak TASKS_PER_ROUND_MULTIPLIER based on the network performance and the charts above, for now manually. We can automate this later if we think it's worth it.

api/lib/round-tracker.js
juliangruber: Yes, I was also thinking about this stateful algorithm, similarly to how you described. Only, I would not operate based on jobs per day, but jobs per round. The desired job count per round is X, depending on whether round n-1 over- or undershot the goal X, adjust the multiplier / offset from round n-1.

bajtos: > Yes, I was also thinking about this stateful algorithm, similarly to how you described. Only, I would not operate based on jobs per day, but jobs per round. The desired job count per round is X, depending on whether round n-1 over- or undershot the goal X, adjust the multiplier / offset from round n-1.

I think that's a good start, your proposal should work well if the number of active Stations is mostly the same every round. Which should be true since most of our network consists of server nodes running 24/7.

However, if there were twice as many nodes running in the UTC evening (America working hours) than in the UTC morning (Asia working hours), then your algorithm will produce more or fewer jobs than we want, depending on whether you calibrate X using the data from UTC morning or evening.

juliangruber: If we calibrate it based on the previous round, then - while we will be 20m behind during a growth/decline phase - we won't be catastrophically off

bajtos: SGTM 👍🏻

juliangruber: Question:

Currently, `spark-stats` operates on `?from` and `&to` filters that are dates or date-time strings.

At first I thought about making the new API `GET /measurements/round?from=roundIndex&to=roundIndex`, but realized I'm fighting against the system constantly.

Therefore I'm wondering whether it's better to have `GET /measurements/per-round?from=dateTime&to=dateTime`, and we track internally at which timestamp a round has started and ended. It's a bit awkward for the query logic (which will live in `spark-api`), as they know the round index, but round start / end times are not easy to synchronize.

From the consumer (`spark-api`) perspective, it would be easiest if the new API supported round index as query filters.

Wdyt @bajtos?

bajtos: Do you need to get values for more than one round? In https://github.com/filecoin-station/spark-api/pull/385#discussion_r1731110859, you proposed to use data from the previous round only. In that case, I think a new endpoint not using any query strings would be the best.

E.g.

```
GET /measurement-count/round/123
```

If you want to make this future proof and support a range of rounds, then yes, let's use round index in the query filters. I propose to use different param names to avoid confusion.

Note that we have a round index that's local to each smart contract version and a round number that's global to the entire Spark deployment.

Based on my past experience, I recommend using the pair (`smartContractAddress, smartContractRoundIndex`) to identify the round you are interested in. That's how "get current round details" works in spark-api.

E.g.

```
GET /measurement-counts?meridianVersion=0x8460766edc62b525fc1fa4d628fc79229dc73031&fromRoundIndex=14800&toRoundIndex=14817
```

Also note that it's super easy to bypass the code handling from/to. Just don't wrap your request handler with `createRespondWithFetchFn` and implement query string parsing and cache TTL yourself.

juliangruber: Agreed, the question above is outdated wrt the comment linked to by you.

What do you think about removing the round index from the query? This way the consumer can be simpler, and it's impossible to create a wrong request. Just:

```
GET /last-round-measurement-count/
```

(Exact path name TBD). This endpoint doesn't need heavy caching, as - for now - we only know of spark-api wanting to use it, every 20 minutes.

> Note that we have a round index that's local to each smart contract version and a round number that's global to the entire Spark deployment.

Removing the round index also resolves this question.

> Also note that it's super easy to bypass the code handling from/to. Just don't wrap your request handler with createRespondWithFetchFn and implement query string parsing and cache TTL yourself.

Very useful point 👍

api/lib/round-tracker.js
```diff
@@ -238,14 +242,30 @@ export async function maybeCreateSparkRound (pgClient, {
...
+      COALESCE(
+        (SELECT max_tasks_per_node FROM spark_rounds WHERE meridian_round = $3 - 1),
+        $5
+      )
+        * $6
+        / COALESCE(
+            (SELECT measurement_count FROM spark_rounds WHERE meridian_round = $3 - 1),
+            $6
+          )
+    )
```
juliangruber, line 260: Verified in https://onecompiler.com/postgresql/42rp72e93:

1. Create round 1, `max_tasks_per_node` will be set to `BASELINE_MAX_TASKS_PER_NODE`
2. Set `measurement_count` to 250k, half of the desired 500k
3. Create round 2, `max_tasks_per_node` will now be double of the previous round.

```
 meridian_round | max_tasks_per_node | measurement_count 
----------------+--------------------+-------------------
              1 |                 15 |            250000
              2 |                 30 |                 0
(2 rows)
```

juliangruber: @bajtos I believe we can refactor this query so it doesn't run two selects on the same table. Otoh, I believe this code path isn't too hot to require this. But, if we would want to do so, @bajtos do you know how to refactor this query?

One option is to move the selects out of this query and into the `if (rowCount)` below.

publish/index.js
```diff
@@ -88,6 +88,18 @@ export const publish = async ({
...
+    await pgClient.query(`
+      UPDATE spark_rounds
+      SET measurement_count = $1
+      WHERE meridian_round = $2
+      ON CONFLICT (measurement_count) DO UPDATE
+      SET measurement_count =
+        spark_round.measurement_count + EXCLUDED.measurement_count
+    `, [
+      measurements.length,
+      roundIndex
```
bajtos, line 97: How do you handle smart-contract upgrades? When we deploy a new smart contract version, the round index resets back to zero.

In round-tracker, we are mapping `(address, roundIndex)` to a monotonically increasing `roundNumber` value. Maybe that's what you can use here?

bajtos: Oh, you are using the existing table `spark_rounds`. In that case, I believe you need to include the contract address in this query.

juliangruber: https://github.com/filecoin-station/spark-api/pull/385/commits/d8d8d5e908f7fbf1148f8a4f36737f5d1b24ed95

api/lib/round-tracker.js
bajtos: Would something like this work?

```sql
WITH prev_round AS (
  SELECT max_tasks_per_node, measurement_count
  FROM spark_rounds WHERE meridian_round = $3 - 1
)
INSERT INTO spark_rounds
  -- (..)
  COALESCE (prev_round.max_task_per_node, $5)
    * $6 
    / COALESCE (prev_round.measurement_count, $6)
```

You may need to convert a the `prev_round` rowset into a scalar, I did something similar recently but don't remember details.

api/lib/round-tracker.js
```diff
@@ -239,43 +237,38 @@ export async function maybeCreateSparkRound (pgClient, {
...
+      $4,
+      COALESCE(
+        (SELECT max_tasks_per_node FROM spark_rounds WHERE meridian_round = $3 - 1),
+        $5
+      )
+        * $6
+        / COALESCE(
+            (SELECT measurement_count FROM spark_rounds WHERE meridian_round = $3 - 1),
+            $6
+          )
```
bajtos: I find this bit extremely difficult to follow. Could you please add SQL comments to explain the formula and what $5 & $6 stands for.

api/lib/round-tracker.js
```diff
@@ -239,43 +237,38 @@ export async function maybeCreateSparkRound (pgClient, {
...
     (id, created_at, meridian_address, meridian_round, start_epoch, max_tasks_per_node)
-    VALUES ($1, now(), $2, $3, $4, $5)
+    VALUES (
+      $1,
+      now(),
+      $2,
+      $3,
+      $4,
+      COALESCE(
+        (SELECT max_tasks_per_node FROM spark_rounds WHERE meridian_round = $3 - 1),
```
bajtos: This isn't going to work when we migrate to a new smart contract version. `meridian_round` will reset to zero.

I think it's safe to use `sparkRoundNumber` instead - the parameter `$1`.

bajtos (changes requested):

api/lib/round-tracker.js
juliangruber: Is this clearer in https://github.com/filecoin-station/spark-api/pull/385/commits/361f9ee2f7f7d1558f1de1ff4036182b13d12ede?

api/lib/round-tracker.js
juliangruber: according to https://stackoverflow.com/a/50364985:

> If the source of an insert statement is a select do not use the VALUES keyword.

But, we do need it here 🤔

juliangruber: It failed in onecompiler as well, saying there's a missing `FROM` clause for `prev_round`

api/lib/round-tracker.js
juliangruber: thanks! https://github.com/filecoin-station/spark-api/pull/385/commits/047eb3f43c65d48e19cfba324444e6f12ab52296

bajtos (changes requested):
Nice! I think the proposed algorithm should work well 👍🏻 

I'd like to improve a few minor details, PTAL below.

api/test/round-tracker.test.js
```diff
@@ -177,49 +263,232 @@ describe('Round Tracker', () => {
...
+                BASELINE_TASKS_PER_NODE * NODE_TASKS_TO_ROUND_TASKS_RATIO
+              )}i`,
+              current_round_node_max_task_count: `${BASELINE_TASKS_PER_NODE}i`,
+              previous_round_measurement_count: '0i',
+              previous_round_node_max_task_count: '0i'
+            }
+          ]
+        )
+      })
+      it('uses baseline values when the previous round was empty', async () => {
```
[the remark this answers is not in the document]
juliangruber: I think we had a convention that style changes need to be enforced by a linter, so they don't need to get discussed. I have followed this convention and really like it. Wdyt?

api/lib/round-tracker.js
```diff
@@ -233,29 +238,72 @@ export async function maybeCreateSparkRound (pgClient, {
...
+      point.intField('previous_round_node_max_task_count', previousRound?.max_tasks_per_node ?? 0)
+    })
   }
 }
 
-async function defineTasksForRound (pgClient, sparkRoundNumber) {
+async function defineTasksForRound (pgClient, sparkRoundNumber, maxTasksPerNode) {
+  const taskCount = Math.floor(
+    maxTasksPerNode * NODE_TASKS_TO_ROUND_TASKS_RATIO
+  )
```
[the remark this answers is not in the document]
juliangruber: Makes sense! https://github.com/filecoin-station/spark-api/pull/385/commits/fdb7513f1c3675662c1a163ed83a470dab1f1757

api/lib/round-tracker.js
```diff
@@ -2,23 +2,24 @@ import assert from 'node:assert'
...
+export const TASKS_EXECUTED_PER_ROUND = 500_000
+
+// Baseline values for how many tasks should be completed every round, and how
+// many tasks each SPARK checker node is expected to complete (every round, at
+// most). The actual value will be set dynamically based on
+// TASKS_EXECUTED_PER_ROUND and the number of tasks executed in the last round.
+export const BASELINE_TASKS_PER_ROUND = 1000
+export const BASELINE_TASKS_PER_NODE = 15
+
+export const NODE_TASKS_TO_ROUND_TASKS_RATIO = BASELINE_TASKS_PER_ROUND / BASELINE_TASKS_PER_NODE
```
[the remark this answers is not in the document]
juliangruber: Thanks for calling this out! That's an important change.

https://github.com/filecoin-station/spark-api/pull/385/commits/c5ab20f2b8aa1460265daf95dc1ea3f21be00bee

api/test/round-tracker.test.js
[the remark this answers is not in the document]
bajtos: I like that convention, too, as long as the linter can enforce things that I find important.

Empty lines used as visual delimiters help me navigate the code faster, so they are important to me.

In some places, like empty lines between test cases, the rule is clear and easy to enforce. In other places, e.g. in the Dockerfile, the decision of where to place empty lines is subjective and difficult to enforce programmatically.

My takeaway is that I should investigate how to incorporate a variant of [this eslint rule](https://github.com/lo1tuma/eslint-plugin-mocha/blob/HEAD/docs/rules/consistent-spacing-between-blocks.md) to our linter checks.

api/lib/round-tracker.js
```diff
@@ -285,11 +285,10 @@ export async function maybeCreateSparkRound (pgClient, {
   if (rowCount) {
     // We created a new SPARK round. Let's define retrieval tasks for this new round.
     // This is a short- to medium-term solution until we move to fully decentralized tasking
```
bajtos, line 287: LOL, this comment did not age that well 😆

bajtos (approved):
:shipit:

api/test/round-tracker.test.js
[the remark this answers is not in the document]
bajtos: Let's continue the discussion here: https://github.com/space-meridian/roadmap/issues/166#issuecomment-2358328908

juliangruber:
Thanks! Will ship tomorrow

PR #387, opened 2024-08-05
feat: add `clients` to each task in round details

bajtos:
For each task selected for retrieval tests, add a new field `clients` with a list of client IDs like `f01234` that are storing the same PieceCID/PayloadCID with the same miner.

Note that a single `(cid, miner_id)` pair can and does identify more than one storage deal.

Links:
- https://github.com/filecoin-station/spark/issues/79


juliangruber (commented):
+1

PR #386, opened 2024-08-05
feat: add client info to retrievable deals

bajtos:
Links:
- https://github.com/filecoin-station/spark/issues/78

PR #388, opened 2024-08-05
feat: get retrievable deals per miner or client

bajtos:
Links:
- https://github.com/filecoin-station/spark-api/pull/386
- https://github.com/filecoin-station/spark/issues/78

Out of the scope of this pull request:

- querying retrievable deals by allocator id

Example request:

```
GET /retrievable-deals/miner/:minerId
```

Example response:

```json
[
  {
    "clientId": "f02516933",
    "dealCount": 6880
  },
  {
    "clientId": "f02833886",
    "dealCount": 3126
  }
]
```

Example request:

```
GET /retrievable-deals/client/:clientId
```

Example response:

```json
[
  {
    "minerId": "f0406478",
    "dealCount": 4592
  },
  {
    "minerId": "f0814049",
    "dealCount": 758
  }
]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,58 @@ Response:
 OK
 ```
 
+### `GET /retrievable-deals/miner/:minerId`
+
+Parameters:
+- `minerId` - a miner id like `f0814049`
+
+Response:
+
+Number of deals grouped by client IDs.
+
+```json
+[
+  {
+    "clientId": "f02516933",
+    "dealCount": 6880
+  },
+  {
+    "clientId": "f02833886",
+    "dealCount": 3126
+  },
+  {
+    "clientId": "f02519404",
+    "dealCount": 3114
+  },
+  {
+    "clientId": "f0215074",
+    "dealCount": 758
+  }
+]
+```
+
+### `GET /retrievable-deals/client/:clientId`
+
+Parameters:
+- `clientId` - a client id like `f0215074`
+
+Response:
+
+Number of deals grouped by client IDs.
+
+```json
+[
+  {
+    "minerId": "f0406478",
+    "dealCount": 4592
+  },
+  {
+    "minerId": "f0814049",
+    "dealCount": 758
+  }
+]
+```
+
 ## Development
 
 ### Database
--- a/api/index.js
+++ b/api/index.js
@@ -28,6 +28,10 @@ const handler = async (req, res, client, domain) => {
     await getMeridianRoundDetails(req, res, client, segs[2], segs[3])
   } else if (segs[0] === 'rounds' && req.method === 'GET') {
     await getRoundDetails(req, res, client, segs[1])
+  } else if (segs[0] === 'retrievable-deals' && segs[1] === 'miner' && req.method === 'GET') {
+    await getRetrievableDealsForMiner(req, res, client, segs[2])
+  } else if (segs[0] === 'retrievable-deals' && segs[1] === 'client' && req.method === 'GET') {
+    await getRetrievableDealsForClient(req, res, client, segs[2])
   } else if (segs[0] === 'inspect-request' && req.method === 'GET') {
     await inspectRequest(req, res)
   } else {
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -58,6 +58,58 @@ Response:
 OK
 ```
 
+### `GET /retrievable-deals/miner/:minerId`
```
patrickwoodhead, line 61: Could we make the endpoint more restful? To me, this is saying that we are fetching objects with the data model given by retrievable deals, and then this is followed by `miner` in the URL which confuses me. And then we get back an array of summary statistics.

Perhaps instead we could go for `GET /deal-stats?groupBy=<miner/client>`

README.md
```diff
@@ -58,6 +58,58 @@ Response:
...
+
+Parameters:
+- `minerId` - a miner id like `f0814049`
+
+Response:
+
+Number of deals grouped by client IDs.
+
+```json
+[
```
patrickwoodhead, line 71: We could wrap the response in 

```json
{ 
    "minerId": "<minerId>",
    "deal-stats": [
        ...
    ]
}
```

README.md
bajtos: I see your point. We need two pieces of information: the field name and the field value.

How about changing the order of path segments as follows:

- `GET /miner/{minerId}/retrievable-deals/summary`
- `GET /client/{clientId}/retrievable-deals/summary`

See [spark-stats/README](https://github.com/filecoin-station/spark-stats/blob/8a3ec03142d5268b96d2846cf2d81efea7c158bc/README.md) for other REST API endpoints we offer, e.g.
- `GET /miners/retrieval-success-rate/summary?from=<day>&to=<day>`

I think this new path structure is very extensible and makes it easy to add various new endpoints:
- stats per allocator id  - `GET /allocator/{id}/retrievable-deals/summary`
- more details about retrievable deals - `GET /miner/{id}/retrievable-deals?page=1&limit=1000`
- etc.

README.md
bajtos: If we adopt my proposal above (`GET /miner/{minerId}/retrievable-deals/summary`), then I propose the following response body format:

```json5
{
  "minerId": "<minerId>",
  "clients": [
    // ...
  ]
}
```

README.md
patrickwoodhead: Yeah this feels better. I like how we are scoping the deals by miner and client. One thing to point out is that I see both `/miners` and `/miner` at different places in the above comment. I personally prefer having plurals so that when you just make a request to `/miners` it returns an array of miners, and then the miner Id afterwards changes it from a list to an get.

bajtos: Makes sense. I believe there are two schools of thought:

1. Use the plural for endpoints concerning multiple instances, use singular when only one instance is in scope. For example:
   - `/pulls` - `https://github.com/filecoin-station/spark-api/pulls`
   - `/pull/388` - `https://github.com/filecoin-station/spark-api/pull/388`
3. Use the plural form everywhere. For example:
   - `/issues` - `https://github.com/filecoin-station/spark/issues`
   - `/issues/78` - `https://github.com/filecoin-station/spark/issues/78`

As you can see, even well-established big companies are not always consistent about this. 😆 

I slightly prefer the first approach, because it avoids the conflict between path segments and entity ids.

For example, if there is an endpoint called `/issues/summary`, then we cannot have an issue that has the ID `summary`.

OTOH, IDs are usually numbers, UUIDs or structured strings like `f0123`, while path segments are usually English words, so the conflict is very unlikely.

However, we are already using the singular form in spark-stats, see e.g. `GET /participant/:address/reward-transfers?from=<day>&to=<day>`

In that light, I prefer consistency and will use the singular form in this repo too.

bajtos:
@patrickwoodhead I reworked the implementation, below are the new examples. I'll need to update the README, but I'll wait until we agree on the new API request & response formats.

Example request:

```
GET /miner/f0814049/retrievable-deals/summary
```

Example response:

```json
{
  "minerId": "f0814049",
  "clients": [
    { "clientId": "f02516933", "dealCount": 6880 },
    { "clientId": "f02833886", "dealCount": 3126 }
  ]
}
```

Example request:

```
GET /clients/f0215074/retrievable-deals/summary
```

Example response:

```json
{
  "clientId": "f0215074",
  "providers": [
    { "minerId": "f0406478", "dealCount": 4592 },
    { "minerId": "f0814049", "dealCount": 758 }
  ]
}
```

bajtos:
```json
{
  "clientId": "f0215074",
  "providers": [
    { "minerId": "f0406478", "dealCount": 4592 },
    { "minerId": "f0814049", "dealCount": 758 }
  ]
}
```

Two questions to consider:

1. Should the top-level property be called `providers` or `miners`?
2. Since the response contains a summary, maybe we can abbreviate the field name `dealCount` to `deals`?

```json
{
  "clientId": "f0215074",
  "miners": [
    { "minerId": "f0406478", "deals": 4592 },
    { "minerId": "f0814049", "deals": 758 }
  ]
}
```

bajtos:
I realised that this new API should be implemented in spark-stats to make it part of the Spark Stats product. I moved the implementation to https://github.com/filecoin-station/spark-stats/pull/194

PR #390, opened 2024-08-07
perf: index retrievable deals by miner & by client

bajtos:
Speed up the queries performed by API returning info about eligible deals.

Links:
- https://github.com/filecoin-station/spark-stats/pull/194
- https://github.com/filecoin-station/spark/issues/78

PR #391, opened 2024-08-07
fix: create index only if not exists

bajtos:
This commit is trying to fix the failed deployment of 49670643.

Links:
- https://github.com/filecoin-station/spark-api/pull/390
- https://github.com/filecoin-station/spark/issues/78


juliangruber (commented):
+1

Issue #404, opened 2024-09-02
Fix Dockerfile to let services handle OS signals


bajtos:
We are invoking the start `CMD` via shell to pass `${SERVICE}` env var in positional arguments

```Dockerfile
CMD npm start --workspace ${SERVICE}
```

The downside of using a shell, quoting from https://docs.docker.com/reference/build-checks/json-args-recommended/:

> When you use shell form, the executable runs as a child process to a shell, which doesn't pass signals. This means that the program running in the container can't detect OS signals like SIGTERM and SIGKILL and respond to them correctly.

We should find a way how to configure npm's workspace via env vars.

Can we perhaps use `NPM_CONFIG_WORKSPACE` for that? It's not as pretty as `SERVICE`, though, as it's leaking the implementation details.

bajtos:
Fixed via #405

PR #405, opened 2024-09-02
chore: fix Docker lint errors

bajtos:
Screenshot 2024-09-02 at 12 03 51

Related:
- https://github.com/filecoin-station/spark-api/issues/404

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,10 +45,7 @@ FROM base
 COPY --from=build /app /app
 
 # Set to `publish` or `api`
-# This argument controls the value passed to npm start --workspace parameter
-ENV SERVICE=""
+# This argument controls the value used by npm to choose which workspace (subdir) to start
+ENV NPM_CONFIG_WORKSPACE=""
 
-# FIXME: don't invoke the service via shell
-# See https://github.com/filecoin-station/spark-api/issues/404
-SHELL ["/bin/bash", "-c"]
-CMD npm start --workspace ${SERVICE}
+CMD [ "npm", "start" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 
 # Adjust NODE_VERSION as desired
 ARG NODE_VERSION=20.12.2
-FROM node:${NODE_VERSION}-slim as base
+FROM node:${NODE_VERSION}-slim AS base
 
 LABEL fly_launch_runtime="nodejs"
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -48,4 +48,7 @@ COPY --from=build /app /app
 # This argument controls the value passed to npm start --workspace parameter
 ENV SERVICE=""
 
+# FIXME: don't invoke the service via shell
+# See https://github.com/filecoin-station/spark-api/issues/404
+SHELL ["/bin/bash", "-c"]
```
juliangruber, line 53: Does this not still use a shell (`/bin/bash`)?

juliangruber (changes requested):

Dockerfile
bajtos: That's a great question! I don't know what shell is executed by default. Maybe `/bin/sh` and perhaps that the one to use instead of Bash?

I was just following the advice given here:
https://docs.docker.com/reference/build-checks/json-args-recommended/

Screenshot 2024-09-02 at 15 37 09

juliangruber: I think this advice says that since you're already using a shell, it's better to be explicit about it

juliangruber: Instead, we should fix the entry point

juliangruber: I think it could be `CMD ['npm', 'start', '--workspace ${SERVICE}']`

bajtos: That won't work; Docker will not resolve the `SERVICE` provided via env var when building the arguments for CMD.

juliangruber: I'm confused, isn't above `${SERVICE}=""`, so we could hardcode that value?

bajtos: I decided to bit the bullet and rework `SERVICE` to `NPM_CONFIG_WORKSPACE`

It seems to work, here is how I verified this change:

```
❯ docker run --rm -it --env=NPM_CONFIG_WORKSPACE=api spark-api

> @filecoin-station/spark-api@0.0.0 start
> node bin/spark.js

Warning: INFLUXDB_TOKEN was not provided by the environment
node:internal/process/esm_loader:34
      internalBinding('errors').triggerUncaughtException(
                                ^

AssertionError [ERR_ASSERTION]: GLIF_TOKEN must be provided in the environment variables
    at file:///app/common/ie-contract-config.js:8:1
    at ModuleJob.run (node:internal/modules/esm/module_job:222:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:323:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
    at async handleMainPromise (node:internal/modules/run_main:113:12) {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}

Node.js v20.12.2
npm ERR! Lifecycle script `start` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: @filecoin-station/spark-api@0.0.0 
npm ERR!   at location: /app/api 
```

juliangruber (approved):
👏 

Better watch the deployment though

PR #409, opened 2024-09-09
fix: accept measurements with `startAt: null`

bajtos:
This is a follow-up for https://github.com/filecoin-station/spark/pull/92


bajtos:
Landing ASAP to prevent spark-api from dropping measurements from checker nodes running the latest Spark version.

PR #423, opened 2024-09-24
publish: add assert logs were emitted. #421

juliangruber:
This will help us solve #421, as
- we have a more semantic error message
- `receipt` is attached to the error, for inspection


bajtos (approved):
:shipit: 

Do you know if the receipt includes the TX hash? If not, then it may be worth adding it to the error as well.

juliangruber:
Let's add it if it doesn't include the info we need

PR #437, opened 2024-10-17
round-tracker: reduce integer overflow risk in multiplication

juliangruber:


api/lib/round-tracker.js
```diff
@@ -271,8 +271,10 @@ export async function maybeCreateSparkRound (pgClient, {
       LEAST(
         $5,
         $6::int /* previousRound.max_tasks_per_node || BASELINE_TASKS_PER_NODE */
-          * $7::int /* TASKS_EXECUTED_PER_ROUND */
-          / $8::int /* previousRound.measurement_count || TASKS_EXECUTED_PER_ROUND */
+          * (
+            $7::int /* TASKS_EXECUTED_PER_ROUND */
+            / $8::int /* previousRound.measurement_count || TASKS_EXECUTED_PER_ROUND */
+          )
```
bajtos, line 277: This change does not look right to me. I understand how this reduces overflow, but it can introduce underflows. I suspect this change explain the outage we saw yesterday - see https://space-meridian.slack.com/archives/C06RPCL6QGL/p1729169875562659

Proposed fix: use a 64-bit integer for the calculations and keep the old older of operations (multiply first, divide second).

Why was it needed to land this PR urgently without waiting for a review?

juliangruber: > Why was it needed to land this PR urgently without waiting for a review?

I pinged you in Slack about this. Please see https://space-meridian.slack.com/archives/C06S76341B8/p1729150814524399 and following. `spark-evaluate` was not able to start new rounds.

juliangruber: > I understand how this reduces overflow, but it can introduce underflows

How can this produce integer unterflow?

> Proposed fix: use a 64-bit integer for the calculations and keep the old older of operations (multiply first, divide second).

I agree, that's the proper fix. I didn't have the time to think about this more. Will follow up.

juliangruber: https://github.com/filecoin-station/spark-api/pull/441

bajtos: When TASKS_EXECUTED_PER_ROUND is smaller than previousRound.measurement_count, then the division yields 0 instead of something like 0.3 or 0.8 we need.

In other words, whenever the network produces too many measurements, we multiply previousRound.max_tasks_per_node by 0.

As I mentioned in the other PR, it would be great to have a test for this.

juliangruber: > When TASKS_EXECUTED_PER_ROUND is smaller than previousRound.measurement_count, then the division yields 0 instead of something like 0.3 or 0.8 we need.

Of course 👍 

> As I mentioned in the other PR, it would be great to have a test for this.

I believe you added a test that covers this here, right? https://github.com/filecoin-station/spark-api/commit/ba375cad391657f82b86fcafdfcf18364023a01a

juliangruber: Hm no we need another test, one where it should calculate `]0, 1[` so we get a smaller value but not 0 or 1

juliangruber: https://github.com/filecoin-station/spark-api/pull/442

PR #438, opened 2024-10-17
round-tracker: add upper limit to max tasks per node

juliangruber:


api/lib/round-tracker.js
```diff
@@ -264,10 +268,11 @@ export async function maybeCreateSparkRound (pgClient, {
       $2,
       $3,
       $4,
-      (
-        $5::int /* previousRound.max_tasks_per_node || BASELINE_TASKS_PER_NODE */
-        * $6::int /* TASKS_EXECUTED_PER_ROUND */
-        / $7::int /* previousRound.measurement_count || TASKS_EXECUTED_PER_ROUND */
+      LEAST(
+        $5,
```
bajtos, line 272: Please add a comment to explain that `$5` is for `MAX_TASKS_PER_NODE_LIMIT`

juliangruber:
> With our 20-minute rounds, 100 tasks per node mean one retrieval check every 12 seconds. That feels a bit too much to me. Having said that, we should be hitting this maximum limit very rarely, so let's see if anybody ever complains.

I thought that every 12 seconds shouldn't be too much for a single node. Do you have another default to suggest?

api/lib/round-tracker.js
juliangruber: https://github.com/filecoin-station/spark-api/commit/58dc884f9f2295edc62ee4ee39beee0bdc6f0ce0

bajtos:
At some point in the past, we used to do one check per minute - that means 20 tasks per node, which feels like a rather low number.

Let's keep the current value and wait for feedback.

PR #439, opened 2024-10-17
remove chainup (fix ci)

juliangruber:
I believe chainup is responsible for the CI errors:

> JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)


bajtos:
Can you improve the error message or the output of CI to make it clear which provider is failing?

juliangruber:
I don't think so, this is thrown from inside Ethers

PR #440, opened 2024-10-17
fix: maxTasksPerNode must be greater than zero

bajtos:
See also:
- https://github.com/filecoin-station/spark-evaluate/pull/382


bajtos:

This was the quickest fix I was able to think about. I would be happy to revert it if you would come up with a more robust solution.

juliangruber (commented):
no looks good!

PR #441, opened 2024-10-18
round-tracker: fix set precision to bigint. #437

juliangruber:
Follow up for #437


bajtos (commented):
The fix LGTM.

I feel it's important to have a test to cover both edge cases (underflow and overflow). Or at the higher level - when we need to significantly increase and decrease the tasks per node calculated for the new round.

juliangruber:
https://github.com/filecoin-station/spark-api/pull/443

Issue #311, opened 2024-05-16
Refactor round tracker to not block service startup


bajtos:
When Glif API performance is degraded, we are not able to deploy new versions of spark-api because the startup times out while fetching RoundStart logs from the chain history.

Let's redesign our code to initialise the round tracker in background and no longer block service startup.

juliangruber:
Adding more context:

- @bajtos: "We can also refactor the startup code to try to check if we are already tracking the current round in the database, I had that in mind, but never found time to implement it."
- I recently increased the block count for the events query from 250 to 500, since spark-evaluate had an issue, and the last event was too old. This blocked spark-api from starting up, as it couldn't find any event. The query shouldn't be too big, as it causes performance degradation from Glif. Idea: add an exponential increase there (check 10 blocks, then 100, etc)

bajtos:
We have already made quite a few improvements in this area. See the list of PRs linked to this issue. 

<img width="417" alt="Image" src="https://github.com/user-attachments/assets/d48e4fd4-9e10-4540-99bb-91c36acf58af" />
<img width="474" alt="Image" src="https://github.com/user-attachments/assets/17dd6fd2-f070-42b5-9a92-5cbf69b1b77e" />

The next step is to assess the current situation, identify the remaining conditions in which the service startup is blocked, and propose an implementation plan.

PR #493, opened 2025-01-19
Add exponential backoff to `getRoundStartEpoch()`. #311

juliangruber:
Ref https://github.com/filecoin-station/spark-api/issues/311#issuecomment-2599838687

> Exponentially look at more blocks to handle the case when we have an outage and the rounds are not advanced frequently enough, while keeping the happy path performant.

### DIFF
--- a/api/lib/round-tracker.js
+++ b/api/lib/round-tracker.js
@@ -331,7 +331,7 @@ async function defineTasksForRound (pgClient, sparkRoundNumber, taskCount) {
 
 // Exponentially look at more blocks to handle the case when we have an outage
 // and the rounds are not advanced frequently enough, while keeping the happy
-// patch performant.
+// path performant.
 export async function getRoundStartEpochWithBackoff (
   contract,
   roundIndex,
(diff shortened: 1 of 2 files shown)


pyropy (approved):
Looks good!! :shipit:

api/lib/round-tracker.js
```diff
@@ -329,17 +329,46 @@ async function defineTasksForRound (pgClient, sparkRoundNumber, taskCount) {
...
+  contract,
+  roundIndex,
+  maxAttempts = 5
+) {
+  for (let attempt = 0; attempt < maxAttempts; attempt++) {
+    try {
+      return await getRoundStartEpoch(
+        contract,
+        roundIndex,
+        50 * Math.pow(2, attempt)
```
bajtos, line 345: Nitpick, feel free to ignore:

```suggestion
        50 * 2 ** attempt
```

Or alternatively to make it clear what is the precedence of operators:

```suggestion
        50 * (2 ** attempt)
```

bajtos (approved):
👏🏻

PR #502, opened 2025-01-23
feat: POST /eligible-deals-batch

bajtos:
See https://github.com/space-meridian/roadmap/issues/218

Create a new API endpoint allowing trusted clients like deal-observer to add
more deals to the list of deals eligible for retrieval testing.

TODO:
- [x] access-token-based authorization

### DIFF
--- a/api/bin/spark.js
+++ b/api/bin/spark.js
@@ -1,4 +1,5 @@
 import '../lib/instrument.js'
+import assert from 'node:assert'
 import http from 'node:http'
 import { once } from 'node:events'
 import { createHandler } from '../index.js'
@@ -13,8 +14,11 @@ const {
   HOST = '127.0.0.1',
   DOMAIN = 'localhost',
   DATABASE_URL,
+  DEAL_INGESTER_TOKEN,
   REQUEST_LOGGING = 'true'
 } = process.env
+
+assert(DEAL_INGESTER_TOKEN, 'DEAL_INGESTER_TOKEN is required')
 
 const client = new pg.Pool({
   connectionString: DATABASE_URL,
@@ -68,6 +72,7 @@ const logger = {
 const handler = await createHandler({
   client,
   logger,
+  dealIngestionAccessToken: DEAL_INGESTER_TOKEN,
   domain: DOMAIN
 })
 const server = http.createServer(handler)
--- a/api/bin/spark.js
+++ b/api/bin/spark.js
@@ -18,6 +18,10 @@ const {
   REQUEST_LOGGING = 'true'
 } = process.env
 
+// This token is used by other Spark services to authenticate requests adding new deals
+// to Spark's database of deals eligible for retrieval testing (`POST /eligible-deals-batch`).
+// In production, the value is configured using Fly.io secrets (`fly secrets`).
+// The same token is configured in Fly.io secrets for the deal-observer service too.
 assert(DEAL_INGESTER_TOKEN, 'DEAL_INGESTER_TOKEN is required')
 
 const client = new pg.Pool({
--- a/api/index.js
+++ b/api/index.js
@@ -36,7 +36,7 @@ const handler = async (req, res, client, domain) => {
     await getSummaryOfEligibleDealsForAllocator(req, res, client, segs[1])
   } else if (segs[0] === 'inspect-request' && req.method === 'GET') {
     await inspectRequest(req, res)
-  } else if (segs[0] === 'ingest-eligible-deals' && req.method === 'POST') {
+  } else if (segs[0] === 'eligible-deals-batch' && req.method === 'POST') {
     await ingestEligibleDeals(req, res, client)
   } else {
     notFound(res)
--- a/api/index.js
+++ b/api/index.js
@@ -9,7 +9,17 @@ import { recordNetworkInfoTelemetry } from '../common/telemetry.js'
 import { satisfies } from 'compare-versions'
 import { ethAddressFromDelegated } from '@glif/filecoin-address'
 
-const handler = async (req, res, client, domain) => {
+/** @import {IncomingMessage, ServerResponse} from 'node:http' */
+/** @import pg from 'pg' */
+
+/*
+ * @param {IncomingMessage} req
+ * @param {ServerResponse} res
+ * @param {pg.Client} client
+ * @param {string} domain
+ * @param {string} dealIngestionAccessToken
+ */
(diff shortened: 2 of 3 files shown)


api/index.js
```diff
@@ -36,6 +36,8 @@ const handler = async (req, res, client, domain) => {
     await getSummaryOfEligibleDealsForAllocator(req, res, client, segs[1])
   } else if (segs[0] === 'inspect-request' && req.method === 'GET') {
     await inspectRequest(req, res)
+  } else if (segs[0] === 'ingest-eligible-deals' && req.method === 'POST') {
```
bajtos, line 39: Discussion point: what's a good endpoint name?

What I have right now:

```
POST /ingest-eligible-deals
```

Alternatives:
- `POST /eligible-deals` (But: such endpoint often expects only one deal, not an array of deals.)

juliangruber: Good reasoning, I would also expect `POST /eligible-deals` to expect one object. Either we make the POSTer perform multiple requests, or I suggest we rename to something like `POST /eligible-deals-bulk`? It's unexpected to have a verb in there

api/test/test.js
```diff
@@ -773,4 +773,88 @@ describe('Routes', () => {
...
+    beforeEach(async () => {
+      await client.query(
+        'DELETE FROM eligible_deals WHERE miner_id = $1',
+        ['f000']
+      )
+    })
+
+    it('ingests new deals', async () => {
+      const deals = [{
+        minerId: 'f000',
```
NikolasHaimerl, line 787: Maybe we want to pull the reused constant values outside of the uni test and make it available for an entire suite of tests? As far as I can tell it is reused throughout the test and this way we avoid errors due to hard coding values over again.

api/index.js
juliangruber: > Have we thought of using something like express.js for the routing, so we do not have that much if else logic?

We're rolling out Fastify atm, see other PRs :)

bajtos: > Also, is there a difference between eligible deals and active deals? Just for consistent naming.

An eligible deal is a deal that Spark considers eligible for retrieval testing.

An active deal is any deal that has not expired yet or was not slashed.

On second thought, the name "active deals" I used in spark-observer may not be the most precise one since spark-observer is tracking active FIL+ deals only. 🤔

api/test/test.js
bajtos: The value is scoped to each test, so using a different string in different tests is okay. It's also not important what exact minerId value we use in these tests, as long as it's a value that does not exist in the real data, so it makes sense to extract it into a shared constant 👍🏻

api/index.js
bajtos: > Are we in risk of reaching the ingress limit here? I think it is usually like 2 or 5 MB.

Great question! What do you mean by the ingress limit? Do you mean the maximum size of the request body allowed by Cloudflare reverse proxy? I don't know what's the limit.

As I see it, it's the responsibility of the client sending there requests (the code in deal-observer) to batch new deals in requests that are small enough to fit into the limit.

In the pull request, the server rejects request with a body larger than 100MB. This can be easily tweaked.

```
const body = await getRawBody(req, { limit: '100mb' })
```

bajtos: Re: endpoint name - I'll use `POST /eligible-deals-batch`

juliangruber: Please don't forget to update the PR name to match

api/bin/spark.js
```diff
@@ -13,9 +14,12 @@ const {
   HOST = '127.0.0.1',
   DOMAIN = 'localhost',
   DATABASE_URL,
+  DEAL_INGESTER_TOKEN,
   REQUEST_LOGGING = 'true'
 } = process.env
 
+assert(DEAL_INGESTER_TOKEN, 'DEAL_INGESTER_TOKEN is required')
```
bajtos, line 21: TODO: generate a random token; set it in fly secrets for spark-api and deal-observer backend.

api/index.js
```diff
@@ -401,7 +411,19 @@ export const inspectRequest = async (req, res) => {
...
+ * @param {IncomingMessage} req
+ * @param {ServerResponse} res
+ * @param {pg.Client} client
+ * @param {string} dealIngestionAccessToken
+ */
+export const ingestEligibleDeals = async (req, res, client, dealIngestionAccessToken) => {
+  if (req.headers.authorization !== `Bearer ${dealIngestionAccessToken}`) {
+    res.statusCode = 403
+    res.end('Unauthorized')
+    return
```
juliangruber, line 424: could be refactored with `status(res, 403)` from `http-responders`, but since we'll switch this to fastify soon, there is little motivation

juliangruber (approved):

api/bin/spark.js
bajtos: Done.

@NikolasHaimerl in the deal-observer backend, the service running on Fly.io can find the access token in the environment variable `DEAL_INGESTER_TOKEN`. (Configured via `fly secrets set`.) We will need to send this token in the Authorization request header when submitting new deals from deals-observer to spark-api.

pyropy: @bajtos Could you please also add a comment above the `DEAL_INGESTER_TOKEN` potentially linking to the comment above describing that token should be set on the `deal-observer` side as well 🙏🏻

api/index.js
```diff
@@ -399,15 +411,73 @@ export const inspectRequest = async (req, res) => {
...
+    INSERT INTO eligible_deals (
+      client_id,
+      miner_id,
+      piece_cid,
+      piece_size,
+      payload_cid,
+      expires_at,
+      sourced_from_f05_state
+    ) VALUES (
+      unnest($1::TEXT[]),
```
pyropy, line 449: There's possibly potential performance improvement here (which we don't have to address now). I have stumbled across this blog post few weeks ago https://www.timescale.com/blog/boosting-postgres-insert-performance

We can do some benchmarks in the future to compare it with current way of how we unnest values.

NikolasHaimerl (approved):

api/bin/spark.js
bajtos: > @bajtos Could you please also add a comment above the `DEAL_INGESTER_TOKEN` potentially linking to the comment above describing that token should be set on the `deal-observer` side as well 🙏🏻

See [cfa2dbd](https://github.com/filecoin-station/spark-api/pull/502/commits/cfa2dbd10d90b6b6f57a6504c6ec074725bba85a)

pyropy (approved):

PR #542, opened 2025-03-03
feat: add allocators to retrieval tasks

NikolasHaimerl:
## Add allocator tracking to retrieval tasks

This PR adds support for tracking allocators in the retrieval tasks system. Now, when tasks are created for a round, we also record which allocators are associated with the clients for each task.

Key changes:
- Added `allocators` array field to retrieval tasks to track all allocators associated with a task
- Updated SQL query in `defineTasksForRound` to collect allocator IDs through a join with the `allocator_clients` table
- Modified the API response in `getMeridianRoundDetails` to include allocators information
- Updated tests to verify allocators array is properly included in task data

This enhancement allows better tracking of which allocators are participating in each task, providing more comprehensive data for our measurement system. This change is necessary to retrieve allocators for clients in the [list of retrieval tasks ](https://github.com/CheckerNetwork/spark-evaluate/blob/3e8a652d8da97f7ba0f31b94dceeafe0826fa545/lib/spark-api.js#L18)inside of `spark-evaluate`.

Related to  https://github.com/CheckerNetwork/roadmap/issues/216


pyropy (approved):
Amazing work Nikolas 👏🏻

api/test/round-tracker.test.js
```diff
@@ -626,6 +626,131 @@ describe('Round Tracker', () => {
             clients: ['f0050']
           }
         ])
+
+        // Clean up
+        await pgClient.query('DELETE FROM retrieval_tasks WHERE round_id = $1', [roundId])
+        await pgClient.query('DELETE FROM spark_rounds WHERE id = $1', [roundId])
+        await pgClient.query("DELETE FROM eligible_deals WHERE miner_id LIKE 'f001%' AND client_id = 'f0050' AND piece_cid LIKE 'baga%' AND piece_size = 1")
```
bajtos, line 633: This block will not be executed when the test fails. (A failed assertion will throw an Error, and the execution of the test function will be aborted at that place.)

Let's move the cleanup to the `after` block above. Alternatively, if you are trying to prevent tests from interfering with each other, then you can do the cleanup in `beforeEach` or `afterEach`.

api/test/round-tracker.test.js
```diff
@@ -626,6 +626,131 @@ describe('Round Tracker', () => {
...
+        await pgClient.query(`
+          INSERT INTO spark_rounds
+          (id, created_at, meridian_address, meridian_round, start_epoch, max_tasks_per_node)
+          VALUES
+          ($1, NOW(), '0x1a', 1, 1, 15)
+        `, [
+          roundNumber
+        ])
+
+        await defineTasksForRound(pgClient, roundNumber, 10)
```
bajtos, line 665: Since you have only three eligible deals to choose retrieval tasks from, calling `defineTasksForRound` with task count 10 may introduce a test failure in the future, if defineTasksForRound starts to check that it produced the right number of tasks.

Let's use task count 3 instead - that's what you want in this test, isn't it?

```suggestion
        await defineTasksForRound(pgClient, roundNumber, 3)
```

The same comment applies to the other test below too.

api/test/round-tracker.test.js
```diff
@@ -626,6 +626,131 @@ describe('Round Tracker', () => {
...
+            allocators: ['allocator1', 'allocator2', 'allocator3']
+          }
+        ])
+
+        // Clean up
+        await pgClient.query('DELETE FROM retrieval_tasks WHERE round_id = $1', [roundNumber])
+        await pgClient.query('DELETE FROM spark_rounds WHERE id = $1', [roundNumber])
+        await pgClient.query("DELETE FROM eligible_deals WHERE miner_id LIKE 'f003%' AND client_id = 'client' AND piece_cid LIKE 'baga%' AND piece_size = 1")
+        await pgClient.query("DELETE FROM allocator_clients WHERE client_id = 'client' AND allocator_id LIKE 'allocator%'")
       })
```
bajtos, line 754: In my first review comment above, I identified a missing test for the case when we choose one deal and need to add clients from other eligible deals to the array of clients for the chosen task.

Please add a similar test for `retrievable_tasks.allocators` too.
- We want multiple eligible deals sharing the `(payload_cid, miner_id)` fields
- Each eligible deal is from a different `client_id`
- Each `client_id` is linked to a different allocator

After running `defineTasksForRound(pgClient, roundNumber, 1)`, the single retrieval task defined should list all clients and all allocators.

bajtos, line 754: Please also add a test to verify that the array `retrieval_tasks.allocators` is de-duplicated. See e122e771513799945f1b733bde84b1732f7e1f44 for inspiration.

The test should fail when the task-building query contains `array_agg(ac.allocator_id) as allocators` and should pass with `array_agg(DISTINCT ac.allocator_id) as allocators`.

bajtos (changes requested):
The implementation goes in the right direction. 👍🏻 

I think there is a subtle bug though, which was not discovered because there was no test for that - see my first comment below.

api/test/round-tracker.test.js
NikolasHaimerl: I intentionally did not put the delete statements into the `beforeEach` or `afterEach` block. The other tests concurrently have access to the databases, and deleting tables in before or after clauses would interfere with them. If I moved the logic where I only delete selected entries in the after or before clauses, then I have to make assumptions about which entries will be created in the tests run between the after and before clauses. I wanted to avoid this logical incoherence by deleting the entries directly in the tests. 

It is not an issue that these statements do not get executed when the test fails. The entries are constructed in a way so they do not interfere with other tests, so whether they are deleted or should not interfere with other tests. I put these deletes in there to clean up the database after the test has finished but only delete those entries which were inserted during a test.

api/lib/round-tracker.js
```diff
@@ -309,18 +309,25 @@ export async function maybeCreateSparkRound (pgClient, {
...
       LIMIT $2
     )
-    SELECT $1 as round_id, selected.payload_cid as cid, selected.miner_id, array_agg(DISTINCT client_id) as clients
+    SELECT 
+      $1 as round_id, 
+      selected.payload_cid as cid, 
+      selected.miner_id, 
+      array_agg(DISTINCT selected.client_id) as clients,
+      array_agg(DISTINCT ac.allocator_id) as allocators
     FROM selected
```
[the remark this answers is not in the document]
NikolasHaimerl: Great catch!

api/test/round-tracker.test.js
NikolasHaimerl: How would you suggest the cleanup in the `after` block would look like? 
From what I saw, we cannot simply delete the entire table as that would interfere with other tests. If we use beforeEach and afterEach then we need to delete the specific entries which is a bit odd. In this case the beforeEach and afterEach blocks would need to know in advance what entries will be inserted during testing.

api/test/round-tracker.test.js
NikolasHaimerl: This test would also cover the testing case for: 

> As the first next step, I would like us to better understand how the current version in main works. Please add a test to the block [describe('defineTasksForRound'](https://github.com/CheckerNetwork/spark-api/blob/7afc388be102d65498fc4f70c520cb2376c24479/api/test/round-tracker.test.js#L566) to demonstrate what happens when eligible_deals contain two entries with the same (payload_cid, miner) fields but different client_id values and we select only one of them for retrieval_tasks.

No?

api/lib/round-tracker.js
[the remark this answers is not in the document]
NikolasHaimerl: Addressed in 59c55d2

api/test/round-tracker.test.js
NikolasHaimerl: An example of what I mean can be found here 59c55d2

api/test/round-tracker.test.js
NikolasHaimerl: @bajtos ping

bajtos: I see two ways to approach this. I prefer the second approach - I find it's more robust, as it does not depend on each test to remember to define cleanup queries for every table touched by that test.

**Option 1**

Create a helper similar to [`defer` keyword in Go](https://www.digitalocean.com/community/tutorials/understanding-defer-in-go).

```js
/** @type {undefined | () => Promise<void>}
let cleanupAfterTest
afterEach(() => {
  if (deferredAfterTest) await deferredAfterTest()
  deferredAfterTest = undefined
}

// in your tests
it('does something', async () => {
  cleanupAfterTest = async () => {
    await pgClient.query('DELETE FROM retrieval_tasks WHERE round_id = $1', [roundId])
    // ...
  })

  // the test code
}
```

**Option 2**

Craft your test data in a way that makes them easy to delete using a generic query.

Insights:

- Tests should be independent. No test should depend on data created by another test. It must be possible to run one test only and the test should still pass.
- Retrieval tasks created by tests have a different expiration time. The tasks we want to preserve have `expires_at = NOW() + INTERVAL '1 year'`. The generic cleanup query can be something like this:
  ```sql
  DELETE FROM eligible_deals WHERE expires < CURRENT_DATE + INTERVAL '1 year'
  ```
  (We cannot use `NOW()` because by the time the test finishes and the cleanup query is executed, `NOW()` returns a later date-time, and so all deals will match the WHERE filter.)
- We are already deleting all retrieval_tasks and spark_rounds before each test, there is no need to clean them up in every test again.
  https://github.com/CheckerNetwork/spark-api/blob/39ff6a1d34dc5284d575a2726ed8a44ca8237dfa/api/test/round-tracker.test.js#L45-L49
- AFAICT, the schema migration scripts don't seed the client-allocator mapping - the table is empty before the test suite starts. It should be fine to delete all entries from that table before each test.
  https://github.com/search?q=repo%3ACheckerNetwork%2Fspark-api%20allocator_clients&type=code
  ```sql
  DELETE FROM allocator_clients;
  ```

  If we needed to preserve some client-allocator mappings, then we can modify tests to always use allocator IDs that will never match real IDs. Real IDs always start with `f0`. You are already using IDs like `allocator` or `allocator1` in your tests. So the cleanup query can be:

  ```sql
  DELETE FROM allocator_clients WHERE NOT starts_with(allocator_id, 'f0');
  ```

api/test/round-tracker.test.js
bajtos: Yes, you are right 👍🏻

bajtos (changes requested):
Almost there!

What's remaining: resolve the discussion about test data cleanup.

api/test/round-tracker.test.js
NikolasHaimerl: I find both options to not be ideal. It requires the user to know a lot about each test and what kind of data is inserted in what place and deleted again. If I write a test and I insert some data, I know which data I inserted. It is thus easy for me to know which data I have to delete after the test again to leave the database in the state that I found it. 

If we were to introduce queries in `before` or `beforeEach` as a developer I will have to know the mechanism of how the tables are maintained. Furthermore, all the tests are running in parallel, so it is quite easy for other tests to influence the test that I am currently trying to write if I do not know exactly how the cleanup mechanism works. 

In genera,l the test setup in `round-tracker.test.js` is quite unintuitive. Each test should either have their own database, which is created and destroyed before and after each test, or each test should have a lock on the database while the test is being executed. Having multiple tests run in parallel against the same database and the same tables is a source of unexpected errors. 

The fact that we need to discuss how to delete data when and where and which data to include and which not is proof to me that the setup is far too complex.

NikolasHaimerl: Addressed in 1fdcb72

api/test/test.js
```diff
@@ -516,6 +516,7 @@ describe('Routes', () => {
         assert.equal(typeof task.cid, 'string', 'all tasks have "cid"')
         assert.equal(typeof task.minerId, 'string', 'all tasks have "minerId"')
         assert(Array.isArray(task.clients), 'all tasks have "clients" array')
+        assert(Array.isArray(task.allocators), 'all tasks have "allocators" array')
         assert(task.clients.length > 0, 'all tasks have at least one item in "clients"')
```
bajtos, line 520: Let's also assert there is at least one allocator for each task.

It's also better to keep assertions grouped by the property they are testing.

```js
        assert(Array.isArray(task.clients), 'all tasks have "clients" array')
        assert(task.clients.length > 0, 'all tasks have at least one item in "clients"')

        assert(Array.isArray(task.allocators), 'all tasks have "allocators" array')
        assert(task.allocators.length > 0, 'all tasks have at least one item in "allocators"')
```

bajtos (approved):
LGTM with one suggestion, no further review is needed from my side.

api/test/test.js
NikolasHaimerl: ac9ebe6

PR #549, opened 2025-03-16
Migrate Spark API service to Fastify 

Goddhi:
This PR migrates the Spark API server from the existing Node.js's built-in HTTP module to Fastify framework.
Changes Made
Added Fastify as a dependency
Created a parallel implementation that maintains compatibility with existing API contracts
Reorganized route handlers

Files worked on 
api/bin/spark-fastify.js - modified this file  for Fastify server entry point 
api/fastify-app.js - Fastify application setup with middleware and hooks
api/routes/index.js - introduced this file for Route definitions migrated from api/index.js, 
api/package.json - Modified this file by adding  Fastify dependency and new start script

### DIFF
--- a/api/bin/spark.js
+++ b/api/bin/spark.js
@@ -1,8 +1,9 @@
 import '../lib/instrument.js'
 import assert from 'node:assert'
-import http from 'node:http'
 import { once } from 'node:events'
-import { createHandler } from '../index.js'
+import Fastify from 'fastify'
+import { createFastifyApp } from '../fastify-app.js'
+// import { createHandler } from '../index.js'
 import pg from 'pg'
 import { startRoundTracker } from '../lib/round-tracker.js'
 import { migrate } from '../../migrations/index.js'
@@ -24,7 +25,7 @@ const {
 // The same token is configured in Fly.io secrets for the deal-observer service too.
 assert(DEAL_INGESTER_TOKEN, 'DEAL_INGESTER_TOKEN is required')
 
-const client = new pg.Pool({
+const pgPool = new pg.Pool({
   connectionString: DATABASE_URL,
   // allow the pool to close all connections and become empty
   min: 0,
@@ -38,20 +39,20 @@ const client = new pg.Pool({
   maxLifetimeSeconds: 60
 })
 
-client.on('error', err => {
+pgPool.on('error', err => {
   // Prevent crashing the process on idle client errors, the pool will recover
   // itself. If all connections are lost, the process will still crash.
   // https://github.com/brianc/node-postgres/issues/1324#issuecomment-308778405
   console.error('An idle client has experienced an error', err.stack)
 })
-await migrate(client)
+await migrate(pgPool)
 
 console.log('Initializing round tracker...')
 const start = new Date()
 
 try {
   const currentRound = await startRoundTracker({
-    pgPool: client,
+    pgPool,
     recordTelemetry: recordNetworkInfoTelemetry
   })
   console.log(
@@ -73,16 +74,21 @@ const logger = {
   request: ['1', 'true'].includes(REQUEST_LOGGING) ? console.info : () => { }
 }
 
-const handler = await createHandler({
-  client,
+// Initialize Fastify app
+const fastifyApp = await createFastifyApp({
+  pgPool,
   logger,
   dealIngestionAccessToken: DEAL_INGESTER_TOKEN,
   domain: DOMAIN
 })
 
-const port = Number(PORT)
-const server = http.createServer(handler)
-console.log('Starting the http server on host %j port %s', HOST, port)
-server.listen(port, HOST)
-await once(server, 'listening')
-console.log(`http://${HOST}:${PORT}`)
+try {
+  await fastifyApp.listen({ 
+    port: Number(PORT), 
+    host: HOST 
+  })
+  console.log(`Server is running on ${fastifyApp.server.address().address}:${fastifyApp.server.address().port}`)
+} catch (err) {
+  console.error('Error starting server:', err)
+  process.exit(1)
+}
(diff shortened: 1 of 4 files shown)


juliangruber:
> npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.

@Goddhi CI is failing because package.json and package-lock.json aren't in sync. Could you please run `npm install` to fix that?

api/bin/spark.js
```diff
@@ -1,8 +1,9 @@
 import '../lib/instrument.js'
 import assert from 'node:assert'
-import http from 'node:http'
 import { once } from 'node:events'
-import { createHandler } from '../index.js'
+import Fastify from 'fastify'
+import { createFastifyApp } from '../fastify-app.js'
+// import { createHandler } from '../index.js'
```
juliangruber, line 6: What is this code comment for?

api/fastify-app.js
```diff
@@ -0,0 +1,70 @@
...
+  dealIngestionAccessToken,
+  domain
+}) {
+  const fastify = Fastify({
+    logger: {
+      level: 'info',
+      transport: {
+        target: 'pino-pretty'
+      }
+    }
```
juliangruber, line 17:
```suggestion
    logger
```

To allow the caller of `createFastifyApp` to provide the logger

api/bin/spark.js
```diff
@@ -73,16 +74,21 @@ const logger = {
   request: ['1', 'true'].includes(REQUEST_LOGGING) ? console.info : () => { }
 }
 
-const handler = await createHandler({
-  client,
+// Initialize Fastify app
+const fastifyApp = await createFastifyApp({
+  pgPool,
```
juliangruber, line 79: please update `logger` according to https://github.com/CheckerNetwork/spark-stats/blob/2558a45d7d17dd10a8671cfe8934ddfd55cf4363/stats/bin/spark-stats.js#L17-L19

api/fastify-app.js
```diff
@@ -0,0 +1,70 @@
...
+  })
```
juliangruber, line 18: Please add https://github.com/CheckerNetwork/spark-stats/blob/2558a45d7d17dd10a8671cfe8934ddfd55cf4363/stats/lib/app.js#L25

api/fastify-app.js
```diff
@@ -0,0 +1,70 @@
...
+
+  // Add shared context to each request
+  fastify.decorateRequest('pgPool', null)
```
juliangruber, line 21: Can we switch to @fastify/postgres instead of manual handling instead please?

api/fastify-app.js
```diff
@@ -0,0 +1,70 @@
...
+  fastify.decorateRequest('dealIngestionAccessToken', null)
+  fastify.decorateRequest('domain', null)
```
juliangruber, line 23: These are constant and don't change from request to request. Therefore, they shouldn't be exposed on the request object.

api/fastify-app.js
```diff
@@ -0,0 +1,70 @@
...
+    request.domain = domain
+
+    // Domain redirect check
+    if (request.headers.host && request.headers.host.split(':')[0] !== domain) {
+      return reply.redirect(301, `https://${domain}${request.url}`)
+    }
+  })
+
+  // Error handler
+  fastify.setErrorHandler((error, request, reply) => {
```
juliangruber, line 38: Can we use fastify's default error handler instead? No need to keep the current implementation, which we only created, because we didn't have any default error handler

api/fastify-app.js
```diff
@@ -0,0 +1,70 @@
...
+      reply.code(500).send('Internal Server Error')
+    }
+
+    if (reply.statusCode >= 500) {
+      Sentry.captureException(error)
+    }
+  })
+
+  // Request logging
+  fastify.addHook('onRequest', (request, reply, done) => {
```
juliangruber, line 54: This can also be removed, as fastify is taking care of request logging now

api/package.json
```diff
@@ -9,6 +9,7 @@
   "scripts": {
     "start": "node bin/spark.js",
     "lint": "standard",
+    "start:fastify": "node bin/spark-fastify.js",
```
juliangruber, line 12:
```suggestion
    "start:fastify": "node bin/spark-fastify.js",
```

That file doesn't exist

api/routes/index.js
juliangruber, line 1: We also need to remove the old routes from https://github.com/CheckerNetwork/spark-api/blob/d3b87c2ae1e7bff084ab9cb36e4c5f3204f0369a/api/index.js

api/routes/index.js
```diff
@@ -0,0 +1,432 @@
...
+    measurement.attestation,
+    inetGroup,
+    measurement.carTooLarge ?? false,
+    measurement.carChecksum,
+    measurement.indexerResult,
+    measurement.minerId,
+    measurement.providerId
+  ])
+  
+  return { id: rows[0].id }
```
juliangruber, line 140:
```suggestion
  reply.send({ id: rows[0].id })
```

juliangruber: I suspect the other routes have similar issues. Once CI is fixed, that should show.

juliangruber (changes requested):

api/fastify-app.js
pyropy: +1 on using `@fastify/postgres`

api/fastify-app.js
Goddhi: Alright, noted

pyropy:
@Goddhi I suggest you link the issue you're working on in the PR description. In this case you should've linked this issue https://github.com/CheckerNetwork/roadmap/issues/220

api/routes/index.js
bajtos, line 1: If we are going to keep all routes in one file, then I prefer to call that file `api/routes.js`.

pyropy (commented):
ping @Goddhi

Hey just wanted to check in and see how things are going. Is there something we could do to help you out?

PR #571, opened 2025-04-09
Add alternative provider retrieval measurement

### LINKED ISSUE #570: Add support for network wide retrievability metric
We new metric that measures how many deals (CIDs) can be retrieved from the network using any available retrieval provider, including non-Filecoin nodes running IPFS.

If a piece is stored with multiple SPs but only one of them serves retrievals, the new score should flag this content as retrievable.

pyropy:
This PR adds support for new alternative-provider check measurement fields added in https://github.com/CheckerNetwork/spark-checker/pull/132. New fields are saved alongside others in the `measurements` and later on published.

Closes #570
Relates to:
- https://github.com/CheckerNetwork/roadmap/issues/254
- https://github.com/CheckerNetwork/spark-checker/pull/132
- https://github.com/CheckerNetwork/spark-evaluate/pull/518


api/index.js
```diff
@@ -124,10 +134,16 @@ const createMeasurement = async (req, res, client) => {
         indexer_result,
         miner_id,
         provider_id,
+        alternative_provider_check_status_code,
+        alternative_provider_check_timeout,
+        alternative_provider_check_car_too_large,
+        alternative_provider_check_end_at,
+        alternative_provider_check_protocol,
+        alternative_provider_check_provider_id,
```
juliangruber, line 142: Throwing this out there: We could alternatively implement this in such a way that after the alternative provider check has been completed, two measurements will have been created. One would link to the other. This would save us from having to duplicate the measurement schema inside itself. I don't think this is worth it yet though

juliangruber: No that suggestion is really not important since spark-api's business is just buffering measurements until it flushes them again. If anything, we should discuss this in a repo that's further down the data processing pipeline

juliangruber (approved):

api/index.js
pyropy: I was thinking about this a bit and you are right. What started of as adding one field (status code for alternative retrieval) turned up into duplicating code a lot. We might be better off with adding a relationship between columns in the measurements table between regular and alternative provider check. That way we could avoid duplicating code down the processing and evaluation pipeline.

api/bin/spark.js
```diff
@@ -27,7 +27,7 @@ assert(DEAL_INGESTER_TOKEN, 'DEAL_INGESTER_TOKEN is required')
 const client = new pg.Pool({
   connectionString: DATABASE_URL,
   // allow the pool to close all connections and become empty
-  min: 0,
+  // min: 0,
```
bajtos, line 30: I am curious - why do we need this change? Can you please add a code comment to capture the reasoning? 

Should we update all other Spark services communicating with PG to use the same settings, too?

api/index.js
```diff
@@ -98,6 +98,16 @@ const createMeasurement = async (req, res, client) => {
   validate(measurement, 'stationId', { type: 'string', required: true })
   assert(measurement.stationId.match(/^[0-9a-fA-F]{88}$/), 400, 'Invalid Station ID')
 
+  if (measurement.alternativeProviderCheck) {
+    validate(measurement, 'alternativeProviderCheck', { type: 'object', required: false })
+    validate(measurement.alternativeProviderCheck, 'statusCode', { type: 'number', required: false })
```
bajtos, line 103: This will produce a confusing error message when the validation fails - the message will mention only `statusCode` but not `alternativeProviderCheck`.

I think the nicest solution would be to improve `validate` to support nested parameters, but that feels like too much work to me.

```suggestion
    validate(measurement, 'alternativeProviderCheck.statusCode', { type: 'number', required: false })
```

Even better solution is to use JSON Schema validation offered by Fastify, but we don't use fastify here yet. Should we take over https://github.com/CheckerNetwork/spark-api/pull/549 to land it first and then continue with this pull request? That does not feel like the fastest way to shipping this feature either :(


How much work would it be to rework the validation to use JSON Schema in this route only, as an interim solution until we land the migration to Fastify?

```js
const MEASUREMENT_SCHEMA = {
  // JSON schema of the measurement
  // We will be able to use this with Fastify later
}

const createMeasurement = async (req, res, client) => {
  const body = await getRawBody(req, { limit: '100kb' })
  const measurement = JSON.parse(body.toString())
  const valid = ajv.validate(schema, data)
  if (!valid) {
    // report errors back to the client 
    // console.log(ajv.errors)
  }
  
  // Some validations cannot be described in JSON Schema
  if (typeof measurement.participantAddress === 'string' && measurement.participantAddress.startsWith('f4')) {
  try {
    measurement.participantAddress = ethAddressFromDelegated(measurement.participantAddress)
  } catch (err) {
    assert.fail(400, 'Invalid .participantAddress - doesn\'t convert to 0x address')
    }
  }

  // etc.
```

Let's discuss!

publish/index.js
```diff
@@ -37,7 +37,13 @@ export const publish = async ({
...
       cid,
       provider_address,
-      protocol
+      protocol,
+      alternative_provider_check_status_code,
+      alternative_provider_check_timeout,
+      alternative_provider_check_car_too_large,
+      alternative_provider_check_end_at,
+      alternative_provider_check_protocol,
+      alternative_provider_check_provider_id
```
bajtos, line 46: Do we have any concerns about the size of JSON measurements uploaded to Storacha? The prefix `alternative_provider_check_` has 28 characters, it's repeated 6x per measurement, and we have ~200K measurements/round - that's 33MB/round to store the prefix.

Have you considered re-building the nested `alternative_provider_check` object instead?

Example measurement to show what I mean:

```json5
{
  "miner": "f01...",
  "provider_id": "somePeerId",
  // ...
  "alternative_provider_check": {
    "provider_id": "someOtherPeerId",
    // ..
  }
}
```

Maybe this is not a concern right now? Let's discuss!

bajtos (changes requested):
The change looks reasonable, I have a few comments to discuss.

api/bin/spark.js
pyropy: Great catch, I must've committed this accidentally.

publish/index.js
pyropy: I haven't been thinking about the size itself. What started as adding single field to the measurement structure quickly turned into adding more and more of them.

I like your suggestion 👍🏻 

Another alternative could be to store a alternative provider check as a separate measurement  with an additional flag. WDYT?

bajtos:
I believe this PR is no longer relevant since we are shutting down the Leaderboard.

PR #687, opened 2025-08-04
chore: fix TypeScript setup & errors

bajtos:


.github/workflows/ci.yml
```diff
@@ -70,6 +70,7 @@ jobs:
           node-version: 22.13.1
       - run: npm ci
       - run: npm run lint
+      - run: npm run test:types
```
pyropy, line 73: I wonder if `lint-all` step should be a requirement for all build steps or could be removed completely as all build steps also test types (by running `npm run test`)? WDYT?

pyropy (approved):

PR #692, opened 2025-08-06
fix: publish empty batches to avoid staled rounds

bajtos:
After we changed the 0k checker node to run all task as quickly as possible (see https://github.com/CheckerNetwork/spark-0k-checker/pull/26), we are hitting a deadlock where no measurements are reported towards the end of the round, and as a result, the round is never advanced.

In this patch, I am changing the publisher to commit a CID for an empty file (`bafkqaaa`) when no measurements were recorded.

### DIFF
--- a/publish/index.js
+++ b/publish/index.js
@@ -44,8 +44,9 @@ export const publish = async ({
     maxMeasurements
   ])
   if (measurements.length === 0) {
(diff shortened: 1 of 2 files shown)


pyropy (approved):
Great job on figuring this out!

PR #693, opened 2025-08-06
feat: set TASKS_PER_ROUND to 500

bajtos:
We are running multiple instances of our 0k checker node, let's make sure they have enough work to do!


pyropy (approved):
:shipit:

